PR #3, opened 2022-03-30
Set packet data to be able to write video/audio packet from caller

aungmoewin:


packet.go
```diff
@@ -30,6 +30,14 @@ func (p *Packet) Data() []byte {
 	})
 }
 
+func (p *Packet) SetData(data []byte) {
+	bytesToC(data, func(b *C.uint8_t, size C.int) error {
+		p.c.data = b
+		p.c.size = size
+		return newError(0)
```
asticode, line 37: you can return `nil` instead

asticode (changes requested):
Could add a simple test [here](https://github.com/asticode/go-astiav/blob/master/packet_test.go#L96) ? A test that would alloc a new packet, use `SetData` then checks data has been set using `Data()`

asticode:
Closing this PR as you should be using [this method](https://github.com/asticode/go-astiav/blob/master/packet.go#L127) instead

Issue #4, opened 2022-04-18
Implementation of Frame.Data() is not correct for all pixel formats


chmgr:
Thanks for the great work you have done with this library!

I tried to use the Frame.Data() method to copy the raw image data of a YUV420P stream and the program sometimes crashed with "unexpected fault address" (segmentation violation).

I tracked the problem down to the algorithm of the Frame.Data method:

```go
func (f *Frame) Data() [NumDataPointers][]byte {
	b := [NumDataPointers][]byte{}
	for i := 0; i < int(NumDataPointers); i++ {
		b[i] = bytesFromC(func(size *C.int) *C.uint8_t {
			*size = f.c.linesize[i]
			if f.c.height > 0 {
				*size = *size * f.c.height
			} else if f.c.channels > 0 {
				*size = *size * f.c.channels
			}
			return f.c.data[i]
		})
	}
	return b
}
```

The YUV420P data consists of three planes. E.g. a 640x480 pixel frame, the planes have the following dimensions:

Y: 640x480
U: 320x240
V: 320x240

However, the implementation always multiplies the `linesize` of the plane with the image height which results in too much bytes being read from C. In the best case, the read image data contains garbage after the correct values, in the worst case, the program tries to read invalid memory addresses.

The AVPixFmtDescriptor struct describes the correct sizes for each plane. However the implementation is much more complicated...

So I would suggest to implement a simple function to receive the raw data of images as simple byte slice:

```go
func (f *Frame) GetImageBuffer() ([]byte, error) {

	// retrieve the image buffer size
	bufferSize := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
	if bufferSize < 0 {
		return nil, newError(bufferSize)
	}

	// create a buffer and copy the raw image data into it
	buffer := make([]byte, int(bufferSize))
	n := C.av_image_copy_to_buffer((*C.uint8_t)(unsafe.Pointer(&buffer[0])), bufferSize, (**C.uint8_t)(unsafe.Pointer(&f.c.data)), (*C.int)(unsafe.Pointer(&f.c.linesize)), C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
	if n < 0 {
		return nil, newError(n)
	}

	return buffer, nil
}
```

[...]

PR #5, opened 2022-04-22
implemented Frame.DataBytes() to receive raw image data

chmgr:
implemented as discussed here #4

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -2,10 +2,12 @@ package astiav
 
 //#cgo pkg-config: libavutil
 //#include <libavutil/frame.h>
+//#include <libavutil/imgutils.h>
 //#include <libavutil/samplefmt.h>
 import "C"
 import (
 	"errors"
+	"fmt"
 	"unsafe"
 )
 
@@ -85,6 +87,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
 		return nil, errors.New("astiav: not implemented")
 	}
 
+}
+
+// SetDataBytes fills the frame data with given raw data. The data must match the pixel format and the dimensions of the frame.
+func (f *Frame) SetDataBytes(data []byte) (err error) {
+
+	// Get the expected buffer length and check data. If data does not have the correct length, we stop and return an
+	// error. Otherwise, pointers will be used to access unexpected memory addresses.
+	expectedBufferLength := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+	if expectedBufferLength < 0 {
+		return newError(expectedBufferLength)
+	}
+	if int(expectedBufferLength) != len(data) {
+		return fmt.Errorf("astiav: invalid data length of %d for frame. Expected %d", len(data), int(expectedBufferLength))
+	}
+
+	// copy data to c
+	buf := (*C.uint8_t)(C.av_malloc(C.size_t(len(data))))
+	if buf == nil {
+		return errors.New("astiav: buf is nil")
+	}
+	C.memcpy(unsafe.Pointer(buf), unsafe.Pointer(&data[0]), C.size_t(len(data)))
+
+	// fill frame with data
+	err = newError(C.av_image_fill_arrays(&f.c.data[0], (*C.int)(&f.c.linesize[0]), buf, (C.enum_AVPixelFormat)(f.c.format), f.c.width, f.c.height, 1))
+	if err != nil {
+		C.av_free(unsafe.Pointer(buf))
+	}
+
+	return
 }
 
 func (f *Frame) Height() int {
--- a/frame.go
+++ b/frame.go
@@ -4,6 +4,10 @@ package astiav
 //#include <libavutil/frame.h>
 //#include <libavutil/samplefmt.h>
 import "C"
+import (
+	"errors"
+	"unsafe"
+)
 
 const NumDataPointers = uint(C.AV_NUM_DATA_POINTERS)
 
@@ -53,6 +57,34 @@ func (f *Frame) Data() [NumDataPointers][]byte {
 		})
 	}
 	return b
+}
+
+// DataBytes returns the frame's data as byte slice.
+func (f *Frame) DataBytes() ([]byte, error) {
+
+	isVideoFrame := f.Height() > 0 && f.Width() > 0
+
+	if isVideoFrame {
+
+		// retrieve the image buffer size
+		bufferSize := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+		if bufferSize < 0 {
+			return nil, newError(bufferSize)
+		}
+
+		// create a buffer and copy the raw image data into it
+		buffer := make([]byte, int(bufferSize))
+		n := C.av_image_copy_to_buffer((*C.uint8_t)(unsafe.Pointer(&buffer[0])), bufferSize, (**C.uint8_t)(unsafe.Pointer(&f.c.data)), (*C.int)(unsafe.Pointer(&f.c.linesize)), C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+		if n < 0 {
+			return nil, newError(n)
+		}
+
+		return buffer, nil
+
+	} else {
+		return nil, errors.New("astiav: not implemented")
+	}
+
 }
 
 func (f *Frame) Height() int {
--- a/frame_test.go
+++ b/frame_test.go
@@ -10,6 +10,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// 4x4 image (all red) in YUV420P format
+var rawYuv420P4x4Red = []byte{0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x51, 0x5a, 0x5a, 0x5a, 0x5a, 0xf0, 0xf0, 0xf0, 0xf0}
+
 func videoInputLastVideoFrame() (f *astiav.Frame, err error) {
 	if global.frame != nil {
 		return global.frame, nil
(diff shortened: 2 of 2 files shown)


asticode (changes requested):
Could you also add a test?

Basically could:

- move `testdata/frame` to `testdata/frame/data`
- update the path [here](https://github.com/asticode/go-astiav/blob/master/frame_test.go#L130) to `testdata/frame/data`
- and replicate a similar test for `DataBytes()` (therefore saving `DataBytes` in a file located at `testdata/frame/data_bytes` and testing `DataBytes` returns the correct data. it seems weird right now since the test has to succeed since we're providing the `expected` and `got` but it will make sense if we want to upgrade `ffmpeg`)

chmgr:
I implemented `SetDataBytes()` as the reverse function to `DataBytes()`. I also added a test verifying that `DataBytes()` returns the same data as set via `SetDataBytes()`. It might be better to check against `Data()` when #4 is fixed.

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
...
+	// error. Otherwise, pointers will be used to access unexpected memory addresses.
+	expectedBufferLength := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+	if expectedBufferLength < 0 {
+		return newError(expectedBufferLength)
+	}
+	if int(expectedBufferLength) != len(data) {
+		return fmt.Errorf("astiav: invalid data length of %d for frame. Expected %d", len(data), int(expectedBufferLength))
+	}
+
+	// copy data to c
```
asticode, line 105: 2 questions:

1) Why do you need to copy the buffer?
2) If you really need to copy the buffer, could you use `bytesToC()` instead ?

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
 
 }
 
+// SetDataBytes fills the frame data with given raw data. The data must match the pixel format and the dimensions of the frame.
+func (f *Frame) SetDataBytes(data []byte) (err error) {
+
```
asticode, line 94: Could you add the `if f.Height() > 0 && f.Width() > 0` logic here as well ?

In fact, could you add a `Frame.isImage() bool` unexported function that returns exactly that, and use it in both `DataBytes` and `SetDataBytes`?

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
...
+	// Get the expected buffer length and check data. If data does not have the correct length, we stop and return an
+	// error. Otherwise, pointers will be used to access unexpected memory addresses.
+	expectedBufferLength := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+	if expectedBufferLength < 0 {
+		return newError(expectedBufferLength)
```
asticode, line 99: Could you wrap the error here ?

`fmt.Errorf("astiav: getting image buffer size failed: %w", newError(bufferSize))`

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
 
 }
 
+// SetDataBytes fills the frame data with given raw data. The data must match the pixel format and the dimensions of the frame.
+func (f *Frame) SetDataBytes(data []byte) (err error) {
+
+	// Get the expected buffer length and check data. If data does not have the correct length, we stop and return an
+	// error. Otherwise, pointers will be used to access unexpected memory addresses.
+	expectedBufferLength := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
```
asticode, line 97: Could your rename `expectedBufferLength` to `bufferSize`?

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
...
+	if expectedBufferLength < 0 {
+		return newError(expectedBufferLength)
+	}
+	if int(expectedBufferLength) != len(data) {
+		return fmt.Errorf("astiav: invalid data length of %d for frame. Expected %d", len(data), int(expectedBufferLength))
```
asticode, line 102: Could you change the error to:

`fmt.Errorf("astiav: invalid buffer size (expected %d, got %d)", int(bufferSize), len(data))`

frame.go
```diff
@@ -87,6 +89,35 @@ func (f *Frame) DataBytes() ([]byte, error) {
...
+
+	// copy data to c
+	buf := (*C.uint8_t)(C.av_malloc(C.size_t(len(data))))
+	if buf == nil {
+		return errors.New("astiav: buf is nil")
+	}
+	C.memcpy(unsafe.Pointer(buf), unsafe.Pointer(&data[0]), C.size_t(len(data)))
+
+	// fill frame with data
+	err = newError(C.av_image_fill_arrays(&f.c.data[0], (*C.int)(&f.c.linesize[0]), buf, (C.enum_AVPixelFormat)(f.c.format), f.c.width, f.c.height, 1))
```
asticode, line 113: Could you wrap the error here?

`fmt.Errorf("astiav: filling image arrays failed: %w", newError(ret))`

frame.go
```diff
@@ -55,6 +61,63 @@ func (f *Frame) Data() [NumDataPointers][]byte {
...
+func (f *Frame) DataBytes() ([]byte, error) {
+
+	isVideoFrame := f.Height() > 0 && f.Width() > 0
+
+	if isVideoFrame {
+
+		// retrieve the image buffer size
+		bufferSize := C.av_image_get_buffer_size(C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+		if bufferSize < 0 {
+			return nil, newError(bufferSize)
```
asticode, line 74: Could you wrap the error here ?

`fmt.Errorf("astiav: getting image buffer size failed: %w", newError(bufferSize))`

frame.go
```diff
@@ -55,6 +61,63 @@ func (f *Frame) Data() [NumDataPointers][]byte {
...
+		}
+
+		// create a buffer and copy the raw image data into it
+		buffer := make([]byte, int(bufferSize))
+		n := C.av_image_copy_to_buffer((*C.uint8_t)(unsafe.Pointer(&buffer[0])), bufferSize, (**C.uint8_t)(unsafe.Pointer(&f.c.data)), (*C.int)(unsafe.Pointer(&f.c.linesize)), C.enum_AVPixelFormat(f.c.format), f.c.width, f.c.height, 1)
+		if n < 0 {
+			return nil, newError(n)
```
asticode, line 81: Could you wrap the error here ?

`fmt.Errorf("astiav: copying image to buffer failed: %w", newError(n))`

frame_test.go
```diff
@@ -124,11 +127,32 @@ func videoInputLastVideoFrame() (f *astiav.Frame, err error) {
 	return
 }
 
+func TestFrame_DataBytes(t *testing.T) {
```
asticode, line 130: 👍

frame_test.go
```diff
@@ -187,4 +211,5 @@ func TestFrame(t *testing.T) {
 	require.NoError(t, err)
 	err = f4.AllocSamples(astiav.SampleFormatS16, 2, 960, 0)
 	require.NoError(t, err)
+
```
asticode, line 214: Could you remove this newline?

frame_test.go
```diff
@@ -124,11 +127,32 @@ func videoInputLastVideoFrame() (f *astiav.Frame, err error) {
...
+		outBytes,
+	)
+}
+
 func TestFrame(t *testing.T) {
 	f1, err := videoInputLastVideoFrame()
 	require.NoError(t, err)
 	b, err := ioutil.ReadFile("testdata/frame")
 	require.NoError(t, err)
+
```
asticode, line 155: Could you remove this newline?

asticode (changes requested):

frame.go
chmgr: 1. the C code sets pointers to this buffer, so i perform a copy to avoid accidental  corruption
2. is bytesToC "safe" concerning the garbage collection? the slice must not be collected after the callback is left. Then we could use it...

chmgr: My C knowledge is very limited, but when I read the code of `av_image_fill_arrays()` which uses `av_image_fill_pointers()` correctly, these functions don't copy the data referenced by the pointer. My conclusion: the memory range of the buffer must be copied to a C managed memory area...

asticode: Thing is, the way I understand it (and I may be wrong), in case of success in your implementation, buffer will never be freed and this will trigger a memory leak (buffer is not freed in C by ffmpeg).

On the other hand, with the implementation I propose, chances are GO GC won't free the buffer until C is not using it anymore. But I'm not 100% sure. I guess this should be tested thoroughly 🤔

chmgr: The more I think about it, `SetDataBytes()` might not be a good idea:

1. Memory management is a problem (I thought that `AVFrame().Free()` would free the buffer, but that is not true)
2. Padding is another problem: data has to be padded with extra bytes depending on resolution and Pixel format (which is not done by `av_image_fill_arrays()`
3. there might be more issues concerning internals of AVFrame

All in all, this seams not to be a good way to load image data to a video frame.

asticode: > Memory management is a problem (I thought that AVFrame().Free() would free the buffer, but that is not true)

Yeah it seems buffer has to be freed separately using `av_free` 🤔 

> Padding is another problem: data has to be padded with extra bytes depending on resolution and Pixel format

Couldn't this be solved by using `av_image_alloc` before using this function?

Nevertheless I tend to agree with you. What do you think about exporting each individual libav functions (for `DataBytes()` as well) ? For instance there would be `(f *Frame) ImageBufferSize()`, `(f *Frame) CopyImageToBuffer()`, `(f *Frame) AllocImage()`, `(f *Frame) FillImageArrays()`, etc.

asticode:
I also had the need to inject raw video inside an encoding workflow and after trying out several solutions, I ended up using [this method](https://github.com/asticode/go-astiav/blob/master/packet.go#L127) (which frees the malloced data once data is not referenced anymore) and then providing the packets to a decoder using codec id `rawvideo`: it manages memory properly and all internal logic such as padding is handled by the decoder. In all honesty, based on the available `libav` API, I even have the feeling that `libav` developers wanted people to do it that way 🤔 

Anyway, thanks for taking the time to write this PR, I'll close it now, but feel free to comment it 👍

PR #6, opened 2022-04-24
implement AllocPacketFromData()

chmgr:
Allow to create packet from byte slice


asticode:
FYI I've implemented a [func (p *Packet) FromData(data []byte) error](https://github.com/asticode/go-astiav/blob/master/packet.go#L127) function therefore I'll close this PR. Still, thanks for taking the time to make this PR ❤️

Issue #8, opened 2022-09-17
Support for sws_scale


elwinar:
Hello there.

I'm trying to do some image manipulation in Golang (things like generating a color palette for a video and timestamps ranges, feeding images to a tensorflow model, etc), and I found out your library while evaluating ffmpeg and linac bindings (I'm an absolute beginer with libav in general.)

Using one of your examples as base, and the conversion function snippet by chmgr here https://github.com/asticode/go-astiav/issues/4#issuecomment-1103204084, I've succeeded in putting each frame in a image.Image.

I have multiple issues with this, the first being that the images seem to have the wrong color (like some green overlay), and the second being that there are a lot more pixel formats than YUV, so my code is essentially limited unless I'm willing to implement such a function for each format.

I've been looking at other examples of how to do this (searching google for anything libav-related seem to always yield ffmpeg or avconv command lines, sic), and finally found out this from the libav-user mailing list https://ffmpeg.org/pipermail/libav-user/2015-September/008473.html.

So, I'm thinking for a first implementation of using sws_scale to convert the frame into RGB and using the result to init an image.RGB, which would have the advantage of supporting all formats, albeit at a performance cost (not necessarily and issue at first).

Does this approach seems correct to you? If so, I didn't see sws functions in your repository, would you be open to a PR implementing the binding?

asticode:
This approach seems pretty good to me if you're aware of the performance cost. An alternative would be to use [this filter](https://ffmpeg.org/ffmpeg-filters.html#format-1) which can convert pixel format as well.

SWS are not in this repository for now indeed, and I'd be open for a PR. In your case you'd need to add 

- a `SWSContext` struct in a `sws_context.go` file
- a `NewSWSContext() *SWSContext` function
- a `func (sc *SWSContext) Scale(...)` function
- tests in a `sws_context_test.go` file

Check out [this context](https://github.com/asticode/go-astiav/blob/master/filter_context.go) so that you can match variable naming, doc and comments.

elwinar:
What I was thinking was to convert to RGB at first, then implement more constructors for other pixel formats as needed and only use the conversion as a fallback for unsupported formats. That way, exotic formats are slower but I can implement them if needed.

I was also thinking, do you think a custom implementation of image.Image that map directly to the frame data could be pertinent? For the supported formats that would alleviate even more the performance penalty of reading from the frames.

asticode:
[...]

PR #10, opened 2022-10-29
Extend IOContext to reader/writer/seeker

BubbaJoe:
## Purpose
Allows you to read/write/seek data that you already have in memory
Allows you to read from a file a file via IOContext


io_context.go
```diff
@@ -3,11 +3,16 @@ package astiav
...
-import "unsafe"
+import (
+	"io"
+	"unsafe"
+)
 
 // https://github.com/FFmpeg/FFmpeg/blob/n5.0/libavformat/avio.h#L161
 type IOContext struct {
 	c *C.struct_AVIOContext
+	io.ReadWriteCloser
```
asticode, line 14: Can you remove those 2 lines?

io_context.go
```diff
@@ -25,15 +30,36 @@ func (ic *IOContext) Closep() error {
...
+		return 0, nil
 	}
 	C.avio_write(ic.c, (*C.uchar)(unsafe.Pointer(&b[0])), C.int(len(b)))
+	return len(b), nil
+}
+
+func (ic *IOContext) Read(b []byte) (n int, err error) {
+	n = int(C.avio_read(ic.c, (*C.uchar)(unsafe.Pointer(&b[0])), C.int(len(b))))
+	if n < 0 {
+		return 0, io.EOF
```
asticode, line 54: I'd rather you return `0, newError(n)` instead

io_context.go
```diff
@@ -25,15 +30,36 @@ func (ic *IOContext) Closep() error {
...
 }
 
-func (ic *IOContext) Write(b []byte) {
+func (ic *IOContext) Write(b []byte) (int, error) {
 	if b == nil {
-		return
+		return 0, nil
 	}
 	C.avio_write(ic.c, (*C.uchar)(unsafe.Pointer(&b[0])), C.int(len(b)))
+	return len(b), nil
```
asticode, line 48: This is misleading since we don't really know if the amount of written bytes is `len(b)` 🤔

io_context.go
```diff
@@ -25,15 +30,36 @@ func (ic *IOContext) Closep() error {
...
+}
+
+func (ic *IOContext) Read(b []byte) (n int, err error) {
+	n = int(C.avio_read(ic.c, (*C.uchar)(unsafe.Pointer(&b[0])), C.int(len(b))))
```
asticode, line 52: Could you not cast this to `int()` right away ?

io_context.go
```diff
@@ -25,15 +30,36 @@ func (ic *IOContext) Closep() error {
...
+	if n < 0 {
+		return 0, io.EOF
+	}
+	return n, nil
```
asticode, line 56: You can cast n to int() here

io_context.go
```diff
@@ -25,15 +30,36 @@ func (ic *IOContext) Closep() error {
...
+}
+
+func (ic *IOContext) Seekable() bool {
+	return int(ic.c.seekable) != 0
+}
+
+func (ic *IOContext) Seek(offset int64, whence int) (int64, error) {
+	return int64(C.avio_seek(ic.c, C.int64_t(offset), C.int(whence))), nil
```
asticode, line 64: Can you use a logic similar to this instead : 

```go

ret := C.avio_seek(ic.c, C.int64_t(offset), C.int(whence))
if ret < 0 {
  return 0, newError(C.int(ret))
}
return int64(ret), nil
```

io_context_test.go
```diff
@@ -1,27 +1,133 @@
...
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/asticode/go-astiav"
 	"github.com/stretchr/testify/require"
 )
 
-func TestIOContext(t *testing.T) {
```
asticode, line 13: As the IOContext tests are not that big, can you avoid splitting them into separate functions?

For instance you could:

- Write to the file using the IOContext => check error + n
- Seek using the IOContext
- Read from the file using the IOContext => check error + n + string == "testing"

Also, could you test `Seekable` as well?

asticode (changes requested):
This is a nice idea event though the `Write` method is a bit misleading 🤔

io_context.go
BubbaJoe: Actually i wanted to use IOContext as an io.Writer / io.Reader.

So i can pass data in memory directly to it.

io_context.go
BubbaJoe: i think `io.Reader` uses io.EOF so that is knows when to stop reading. Let me try it out though.

io_context_test.go
BubbaJoe: Sure, but concerning `Seekable`; im not sure if what case it would be false. Let me know if you have any suggestions for that.

io_context.go
BubbaJoe: Yea, i realized this too while coding it, but didn't think it was a big deal. I think there might be a way to tell but i dont quite understand the C code here.

https://ffmpeg.org/doxygen/5.0/aviobuf_8c_source.html#l00232

io_context.go
asticode: Sure, but there's no need to declare interfaces here, the simple fact that IOContext would implement those interfaces is sufficient

io_context_test.go
asticode: I think testing it's `true` when opening a file is sufficient 👍

io_context.go
asticode: Thing is, I want `astiav` to be as close as possible to C since it's merely an expression of the C code. Implementing `io.ReadWriteCloser` as well as the `io.Seeker` interfaces seems like a good idea but I'm wondering whether we should create a wrapper on top of `IOContext` instead 🤔 

Things that are bothering me:

- the `Close` method that would implement the `io.Closer` interface should use `Closep()` instead
- the `Write` method shouldn't return `(n int)` since C doesn't provide that information
- the `Read` method should return an `astiav` error

Here's what I'm proposing:

- update `IOContext` methods you've added so that they're closer to C (cf. my other comments)
- create an `IOContextWrapper` struct (let me know if you think of a better name) that would only implement the `io.ReadWriteCloser` as well as the `io.Seeker` interfaces (in those methods, you could return `io.EOF` for instance)
- add an `IOContext`.`Wrapper()` method that would return an `IOContextWrapper`

That people that want to stay close to C can use `IOContext`, and people that want to use GO interfaces can use `IOContextWrapper`

BubbaJoe: Thanks for the proposal. I am currently looking at https://ffmpeg.org/doxygen/5.0/structAVIOContext.html

And it seems that there are some fields that I can use to extract information about how many bytes were written; testing in progress.

Concerning, the `io.Closer`, i actually just realized that we probably shouldn't implement it as `avio_closep(...)` (and `avio_close(...)`) is only needed when the resource is opened by`avio_open(...)`

I can manually check [eof_reached](https://ffmpeg.org/doxygen/5.0/structAVIOContext.html#aa9214f54c01f07076e0a0718b2e9c27b), and return `io.EOF` as the error.

What do you think? Do you still think a wrapper is better? I think the API will be confusing/weird with wrappers.

FYI: [Read needs to implement io.EOF](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/io/io.go;l=44)

asticode: > I think the API will be confusing/weird with wrappers.

I agree with you.

I don't think implementing the `io.ReadWriteCloser` and `io.Seeker` is a good idea in this lib since its main mission is to be the closest possible to C. If somebody wants to implement this interface, he/she can do it in their project, it's pretty straight-forward.

Therefore here's what I'd rather have in this PR:

- add `func (ic *IOContext) Close() error`
- keep `Write()` as it is
- add `func (ic *IOContext) Read(b []byte) (n int, err error)`
- add `func (ic *IOContext) Seekable() bool`
- add `func (ic *IOContext) Seek(offset int64, whence int) (int64, error)`
- add `IOContext` getters if you need them to implement the `io` interfaces (such as `bytes_written` for instance)

and that's it. That should provide the proper method to anybody looking to implement `io.ReadWriteCloser` and `io.Seeker`.

BubbaJoe: Makes sense will do.

asticode:
You definitely need to alloc the context manually if you don't use `Open()`. But that's a pretty hard change to do with callbacks and free() constraints. I wouldn't do it if I were you and stick with Open() for testing

BubbaJoe:
Hello, Sorry for the delay; got distracted by all of the new games 😵‍💫,

After your comments i realized that i was implementing everything wrong. So i re-did everything from the start. The callback functionality needed a bit of hacking to get working, but it looks 10 times cleaner than my first iteration of it. Anyways, I think this implementation is idiomatic to work seamlessly with go standards, but not too much because you can still use the original C functions directly.

One issue that I am still having is that once I create the IOContext using `avio_alloc_context(...)`; I cannot use `(*IOContext) Close()` or  `(*IOContext) Closep()` to close, as it results in a segmentation violation. However, `(*IOContext) Free()` is working and i am using that in tests. I am thinking of using `(*IOContext) Free()` when `(*IOContext) Close()` is called in the case that i use `avio_alloc_context`.

Please try out the tests i added:
go test -benchmem -run ^TestIOContext -v 

Please try out the benchmarks i added:
`go test -benchmem -run ^$ -bench '^BenchmarkIOContext' -cpu 1,2,4,8,16`


Additionally, I needed support for sws_scale (#8), so I implemented this but i will put it in another PR as soon I add proper testing.

I have created a cool (gocv example)[https://github.com/BubbaJoe/go-astiav-contr/tree/misc-update/examples/gocv], let me know if you are interesting in having examples that are different from original ffmpeg examples. Thanks!

BubbaJoe:
@asticode

PR #29, opened 2023-07-19
Add avdevice support

Tryanks:
Add avdevice support to turn on local cameras or other video devices.
Already tested with v4l2 on Linux, works great.
Example code:
```go
	astiav.RegisterAllDevices()
	dictionary := astiav.NewDictionary()
	dictionary.Set("pixel_format", "mjpeg", 0)
	dictionary.Set("framerate", "30", 0)
	dictionary.Set("video_size", "1920x1080", 0)
	inputFormat := astiav.FindInputFormat("video4linux2")
	formatContext := astiav.AllocFormatContext()
	err := formatContext.OpenInput("/dev/video0", inputFormat, dictionary)
	if err != nil {
		panic(err)
	}
	err = formatContext.FindStreamInfo(nil)
	if err != nil {
		panic(err)
	}
	pkt := astiav.AllocPacket()
	for {
		err = formatContext.ReadFrame(pkt)
		if err != nil {
			panic(err)
		}
		// jpeg data per frame
		data := pkt.Data()
		fmt.Println(len(data))
	}
```

### DIFF
--- a/input_format_test.go
+++ b/input_format_test.go
@@ -1,0 +1,12 @@
+package astiav_test
+
+import (
+	"github.com/asticode/go-astiav"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestFindInputFormat(t *testing.T) {
+	inputFormat := astiav.FindInputFormat("video4linux2")
+	require.True(t, inputFormat.Flags().Has(astiav.IOFormatFlagNoByteSeek))
+}
--- a/input_format_test.go
+++ b/input_format_test.go
@@ -6,7 +6,9 @@ import (
 	"testing"
 )
 
-func TestFindInputFormat(t *testing.T) {
-	inputFormat := astiav.FindInputFormat("video4linux2")
-	require.True(t, inputFormat.Flags().Has(astiav.IOFormatFlagNoByteSeek))
+func TestInputFormat(t *testing.T) {
+	formatName := "rawvideo"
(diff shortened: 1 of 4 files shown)


input_format_test.go
```diff
@@ -0,0 +1,12 @@
+package astiav_test
+
+import (
+	"github.com/asticode/go-astiav"
+	"github.com/stretchr/testify/require"
+	"testing"
+)
+
+func TestFindInputFormat(t *testing.T) {
```
asticode, line 9: Could you:

- add both `func (f *InputFormat) Name() string` and `func (f *InputFormat) String() string` functions to the `InputFormat` struct, similar to the `Codec` struct?
- use the `rawvideo` input format instead in the test (otherwise test fails in windows or darwin)
- check first `inputFormat` is not nil and then use the `.Name()` function instead of the `.Flags()` function in the test

asticode (changes requested):

input_format_test.go
asticode, line 9: Could you rename this test to `TestInputFormat` instead?

Tryanks:
Thanks for the suggested changes.
I didn't realize: the addition of `av_find_input_format` to support avdevice is a major change.
I think there may be a need to add a series of peripheral features and unit tests around `FindInputFormat`?

asticode:
Thanks for the PR! ❤️

Issue #36, opened 2024-01-23
Reusing golang image when use framedata image


Cacsjep:
I'm using go-astiav to decode frames for a CCTV camera client, integrating with Fyne for the UI rendering. The renderer in this setup requires image.RGBA format from Go's standard image library. Given the scale of my application – 16 cameras at 25 frames per second – there's a significant volume of images being created in the process of decoding and use **framedata.Image()**. 

The FrameData.Image() method provided by go-astiav is quite effective, but it generates a new image instance with each invocation. 
This high throughput of image data introduces substantial workload for Go's garbage collector (GC). While this functionality is generally useful, in my context, where performance and resource efficiency are critical, it becomes less ideal. To address this, I've developed a small func that fits for my need.

Are u open for a PR for this, but i have no really deep understanding of image formats so its hard for me if you want that this should work for all image formats.
```go
func (d *FrameData) CopyIntoRgbaImage(dstImage image.RGBA) (image.RGBA, error) {
	if d.f.PixelFormat() == PixelFormatRgba {
		if d.f == nil || d.f.Width() <= 0 || d.f.Height() <= 0 {
			return dstImage, errors.New("invalid frame: either nil or has non-positive dimensions")
		}
		copy(dstImage.Pix, d.planeData(0, func(linesize int) int { return linesize * d.f.Height() }))
		dstImage.Stride = d.f.Linesize()[0]
		dstImage.Rect = image.Rect(0, 0, d.f.Width(), d.f.Height())
		return dstImage, nil
	}
	return dstImage, errors.New("invalid frame: its not an rgba frame")
}
```

btw awesome work with this lib =)

Cacsjep:
I create a PR for just doing planeData public so devs could create functions like i did without any big change to the codebase.

Cacsjep:
pr open

asticode:
Actually I quite liked the idea of being able to chose between creating a new image or updating an existing one therefore I've just added a `FrameData`.`ToImage(i image.Image)` to `master`. Can you tell me whether that fixes your problem?

Cacsjep:
cool, looks very interesting. for me i need to update a image.RGBA and not an image.NRGBA it would be greate if you could add this to ;) then it would be perfect for me =)

asticode:
Thing is I had problems in the past using `image.RGBA` with image not being encoded or displayed properly therefore that's why I now only use `image.NRGBA`

Cacsjep:
[...]

PR #37, opened 2024-01-23
Make planedata public

Cacsjep:
The proposed change involves making the planeData method public, thereby allowing direct access to the underlying frame data.

This change would be bringing the possibility to create such functions like this below.

```go
func CopyIntoRgbaImage(f astiav.Frame, dstImage *image.RGBA) {
    requiredSize := f.Linesize()[0] * f.Height()
    
    // If the existing Pix slice is smaller than needed, allocate a new slice
    if len(dstImage.Pix) < requiredSize {
        dstImage.Pix = make([]byte, requiredSize)
    }

    // Copy pixel data
    copy(dstImage.Pix, f.Data().PlaneData(0, func(linesize int) int { return linesize * f.Height() }))

    // Update stride if different
    if dstImage.Stride != f.Linesize()[0] {
        dstImage.Stride = f.Linesize()[0]
    }

    // Update rect if different
    if dstImage.Rect.Dx() != f.Width() || dstImage.Rect.Dy() != f.Height() {
        dstImage.Rect = image.Rect(0, 0, f.Width(), f.Height())
    }
}
```

### DIFF
--- a/frame_data.go
+++ b/frame_data.go
@@ -44,7 +44,7 @@ func (d *FrameData) Bytes(align int) ([]byte, error) {
 	return nil, errors.New("astiav: frame type not implemented")
 }
 
-func (d *FrameData) planeData(i int, sizeFunc func(linesize int) int) []byte {
+func (d *FrameData) PlaneData(i int, sizeFunc func(linesize int) int) []byte {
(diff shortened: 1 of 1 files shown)


asticode:
Closing this PR

See https://github.com/asticode/go-astiav/issues/36

PR #39, opened 2024-01-26
Added Q2D + GetTimeRelative

dawn9527:
Usually we need the implementation of these two functions,Please review！
Thanks!

### DIFF
--- a/time.go
+++ b/time.go
@@ -12,3 +12,11 @@ const (
 var (
 	TimeBaseQ = newRationalFromC(C.AV_TIME_BASE_Q)
 )
+
+func Q2D(a Rational)  float64{
+	return float64(C.av_q2d(a.c))
+}
+
+func GetTimeRelative()  int64{
+	return int64(C.av_gettime_relative())
+}


time.go
```diff
@@ -12,3 +12,11 @@ const (
 var (
 	TimeBaseQ = newRationalFromC(C.AV_TIME_BASE_Q)
 )
+
+func Q2D(a Rational)  float64{
```
asticode, line 16: What is the difference between [this function](https://github.com/asticode/go-astiav/blob/master/rational.go#L40) and the function you've added? 🤔  (I've renamed the function in `go-astiav` from `ToDouble` to `Float64` since that may be clearer)

time.go
```diff
@@ -12,3 +12,11 @@ const (
...
+	return float64(C.av_q2d(a.c))
+}
+
+func GetTimeRelative()  int64{
```
asticode, line 20: Could you rename this function to `RelativeTime`?

asticode (changes requested):
Thanks for the PR 👍

time.go
dawn9527: Your approach looks better！:satisfied:

PR #40, opened 2024-01-28
Enhance Codec Iteration and Checking of HardwareConfigMethodFlags

Cacsjep:
Hey 

During intergration steps of hw decoding I see that some thing would be very handy.

**1. Hardware Configuration Flag Check:**
 A method HasHardwareConfigMethodFlag in the Codec struct allows checking if a codec supports a specific hardware configuration method. This is particularly useful for identifying if a codec is hardware accelerated.

Implementation snippet:
```go
func (c *Codec) HasHardwareConfigMethodFlag(chcmf CodecHardwareConfigMethodFlag) bool {
    var i C.int
    for {
        config := C.avcodec_get_hw_config(c.c, i)
        if config == nil {
            break
        }
        if (CodecHardwareConfig{c: config}.MethodFlags().Has(chcmf)) {
            return true
        }
        i++
    }
    return false
}
```

**2. Codec Iteration Enhancement:** 
A new function IterateCodecs simplifies the process of iterating over available codecs. This function takes a CodecProcessor and applies it to each codec, facilitating custom processing logic per codec.

Implementation snippet:
```go
func IterateCodecs(processor CodecProcessor) {
    var opq *C.void = nil
    for {
        c := C.av_codec_iterate((*unsafe.Pointer)(unsafe.Pointer(&opq)))
        if c == nil {
            break
        }
        codec := newCodecFromC(c)
        processor(codec)
    }
}
```

**3. Codec Identification:**
Added a method ID to the Codec struct to retrieve the codec's ID, aiding in the identification and filtering of codecs.
```go
func (c *Codec) ID() CodecID {
    return CodecID(c.c.id)
}
```

**Use Case**
These enhancements allow for efficient identification and selection of codecs based on capabilities and specific codec IDs. This is particularly useful in scenarios where an application needs to support multiple codecs across different operating systems and offers the user the ability to switch between different decoders.

A practical example of using these enhancements is shown below:
```go
var decoders []*astiav.Codec
processor := func(c *astiav.Codec) {
    if c.IsDecoder() && c.HasHardwareConfigMethodFlag(astiav.CodecHardwareConfigMethodFlagHwDeviceCtx) && c.ID() == astiav.CodecIDMjpeg {
        decoders = append(decoders, c)
    }
}
astiav.IterateCodecs(processor)
```

Br christoph

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -13,6 +13,8 @@ type Codec struct {
 	c *C.struct_AVCodec
 }
 
+type CodecProcessor func(*Codec)
+
 func newCodecFromC(c *C.struct_AVCodec) *Codec {
 	if c == nil {
 		return nil
@@ -26,6 +28,10 @@ func (c *Codec) Name() string {
 
 func (c *Codec) String() string {
 	return c.Name()
+}
+
+func (c *Codec) ID() CodecID {
+	return CodecID(c.c.id)
 }
 
 func (c *Codec) ChannelLayouts() (o []ChannelLayout) {
@@ -112,4 +118,34 @@ func (c *Codec) HardwareConfigs(dt HardwareDeviceType) (configs []CodecHardwareC
 		i++
 	}
 	return
+}
+
+func (c *Codec) HasHardwareConfigMethodFlag(chcmf CodecHardwareConfigMethodFlag) bool {
+	var i C.int
+	for {
+		config := C.avcodec_get_hw_config(c.c, i)
+		if config == nil {
+			break
+		}
+
+		if (CodecHardwareConfig{c: config}.MethodFlags().Has(chcmf)) {
+			return true
+		}
+
+		i++
+	}
+	return false
+}
+
+func IterateCodecs(processor CodecProcessor) {
+	var opq *C.void = nil
+	for {
+		c := C.av_codec_iterate((*unsafe.Pointer)(unsafe.Pointer(&opq)))
+		if c == nil {
+			break
+		}
+
+		codec := newCodecFromC(c)
+		processor(codec)
+	}
 }
--- a/codec.go
+++ b/codec.go
@@ -107,7 +107,7 @@ func FindEncoderByName(n string) *Codec {
 	return newCodecFromC(C.avcodec_find_encoder_by_name(cn))
 }
 
-func (c *Codec) HardwareConfigs(dt HardwareDeviceType) (configs []CodecHardwareConfig) {
+func (c *Codec) HardwareConfigs() (configs []CodecHardwareConfig) {
 	var i int
 	for {
 		config := C.avcodec_get_hw_config(c.c, C.int(i))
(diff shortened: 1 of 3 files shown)


codec.go
```diff
@@ -113,3 +119,33 @@ func (c *Codec) HardwareConfigs(dt HardwareDeviceType) (configs []CodecHardwareC
 	}
 	return
 }
+
+func (c *Codec) HasHardwareConfigMethodFlag(chcmf CodecHardwareConfigMethodFlag) bool {
```
asticode, line 123: I'd rather not have this function in the lib, it seems this logic could be achieved with `(c *Codec).HardwareConfigs()` with relatively little lines of codes and is present in the `hardware decoding` example

codec.go
```diff
@@ -113,3 +119,33 @@ func (c *Codec) HardwareConfigs(dt HardwareDeviceType) (configs []CodecHardwareC
...
+		if (CodecHardwareConfig{c: config}.MethodFlags().Has(chcmf)) {
+			return true
+		}
+
+		i++
+	}
+	return false
+}
+
+func IterateCodecs(processor CodecProcessor) {
```
asticode, line 140: I'd rather have a global `Codecs() []*Codec` method returning all available codecs

codec_test.go
```diff
@@ -66,4 +66,13 @@ func TestCodec(t *testing.T) {
```
asticode, line 66: Line` 13` could you add a test for `c.ID()`?

codec_test.go
```diff
@@ -66,4 +66,13 @@ func TestCodec(t *testing.T) {
 
 	c = astiav.FindDecoderByName("invalid")
 	require.Nil(t, c)
+
+	var decoders []*astiav.Codec
```
asticode, line 70: I'd replace all this with
```go
var found bool
for _, c := range astiav.Codecs() {
  if c.ID() == astiav.CodecIDMjpeg {
    found = true
  }
}
require.True(t, found)
```

I'd even be tempted to test the whole content of the slice, but I'm not sure windows, linux and macosx will return the same list 🤔  I'll check it out when this PR will be merged 👍

asticode (changes requested):
Thanks for the suggestions ! ❤️

codec.go
Cacsjep: ok but would u think its not better to have also av_codec_iterate in the lib

codec_test.go
Cacsjep: sure

codec.go
Cacsjep: Oh i see there is an issue with "HardwareConfigs" because the devicetype is never used. We should remove the Argument and Update the hw example

codec_test.go
Cacsjep: Thats true it depense on ffmpeg build it could be that ffmpeg is build without specific Codecs, I See that u use mjpeg for Testing, so thats was the reason why i did this to. We also could use h246 i think most builds would have this.

codec.go
Cacsjep: I mean at end it would be the Same :)

Cacsjep:
Done =)

asticode:
Perfect, I'll merge the PR and make minor adjustments 👍

PR #49, opened 2024-02-27
Add string representation to log level

leandromoreira:


log.go
```diff
@@ -37,6 +37,11 @@ import "unsafe"
 
 type LogLevel int
 
+// https://github.com/FFmpeg/FFmpeg/blob/release/5.1/libavutil/log.c#L268
+func (l LogLevel) String() {
```
asticode, line 41: Could you move this function below constants?

leandromoreira:
I see `char *get_level_str(int level)` is not exposed through log.h

leandromoreira:
It seems it won't work since is not exposed :( 
```
log.cover.cgo2.c:(.text+0x25e): undefined reference to `get_level_str'
```

leandromoreira:
I'll close it for now.

PR #52, opened 2024-03-11
Add setter and getter for extra_hw_frames

l0rem1psum:
Added this because I need to use it :)

### DIFF
--- a/codec_context.go
+++ b/codec_context.go
@@ -314,6 +314,14 @@ func (cc *CodecContext) SetHardwareDeviceContext(hdc *HardwareDeviceContext) {
 	}
 }
 
+func (cc *CodecContext) ExtraHardwareFrames() int {
+	return int(cc.c.extra_hw_frames)
+}
+
+func (cc *CodecContext) SetExtraHardwareFrames(n int) {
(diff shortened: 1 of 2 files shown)


codec_context.go
```diff
@@ -314,6 +314,14 @@ func (cc *CodecContext) SetHardwareDeviceContext(hdc *HardwareDeviceContext) {
 	}
 }
 
+func (cc *CodecContext) ExtraHwFrames() int {
```
asticode, line 317: Could you rename to `ExtraHardwareFrames`?

codec_context.go
```diff
@@ -314,6 +314,14 @@ func (cc *CodecContext) SetHardwareDeviceContext(hdc *HardwareDeviceContext) {
...
+	return int(cc.c.extra_hw_frames)
+}
+
+func (cc *CodecContext) SetExtraHwFrames(n int) {
```
asticode, line 321: Could you rename to `SetExtraHardwareFrames`?

codec_context.go
l0rem1psum: Done

Issue #53, opened 2024-03-14
Return `unsafe.Pointer` instead of raw C type


l0rem1psum:
As discussed in a previous PR https://github.com/asticode/go-astiav/pull/44, the raw C struct of `AVFrame` is returned instead of a `unsafe.Pointer`.

However, after some fiddling around, I finally discovered the difference of the two approaches.

Currently, the raw C struct returned will have a type of `*astiav._Ctype_struct_AVFrame`, and as the name suggests, this type is unexported. Hence, it comes with all the inconveniences of returning an unexported type.

To make it usable, one unavoidably has to do the following:

![image](https://github.com/asticode/go-astiav/assets/28675546/b47063b3-6073-41ad-82de-473d6f995467)

In conclusion, it's better to just return `unsafe.Pointer` and let the caller do the casting.

asticode:
Thanks for the detailed explanation 👍 The best would be to replace `UnsafeTypedPointer` to `UnsafePointer` which would indeed return an `unsafe.Pointer`. Do you feel like creating a PR for this?

PR #54, opened 2024-03-14
Return unsafe pointer instead of raw C struct

l0rem1psum:
This addresses the issue: https://github.com/asticode/go-astiav/issues/53

### DIFF
--- a/frame_test.go
+++ b/frame_test.go
@@ -3,6 +3,7 @@ package astiav
 import (
 	"bytes"
 	"testing"
+	"unsafe"
 
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


frame_test.go
```diff
@@ -12,7 +12,7 @@ func TestFrame(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, [8]int{384, 192, 192, 0, 0, 0, 0, 0}, f1.Linesize())
 	require.Equal(t, int64(60928), f1.PktDts())
-	require.Equal(t, f1.c, f1.UnsafeTypedPointer())
+	require.NotNil(t, f1.UnsafePointer())
```
asticode, line 15: Does it work if you use `require.Equal(t, unsafe.Pointer(f1.c), f1.UnsafePointer())` instead? 🤔

l0rem1psum: Fixed and all tests passed

PR #56, opened 2024-03-29
add SetExtraData to CodecParameters to set extradata 

Cacsjep:
For publishing an rtmp stream extradata needs to be set (PPS, SPS h264).

### DIFF
--- a/codec_parameters.go
+++ b/codec_parameters.go
@@ -145,6 +145,17 @@ func (cp *CodecParameters) SetSampleAspectRatio(r Rational) {
 	cp.c.sample_aspect_ratio = r.c
 }
 
+func (cp *CodecParameters) ExtraData() []byte {
+	return bytesFromC(func(size *C.size_t) *C.uint8_t {
+		if cp.c.extradata == nil {
+			*size = C.size_t(0)
+			return nil
+		}
+		*size = C.size_t(cp.c.extradata_size)
+		return cp.c.extradata
+	})
+}
+
 func (cp *CodecParameters) SetExtraData(extraData []byte) error {
 	if len(extraData) == 0 {
 		return nil
@@ -156,9 +167,8 @@ func (cp *CodecParameters) SetExtraData(extraData []byte) error {
 	}
 
 	extradataSize := len(extraData)
-	cp.c.extradata = (*C.uint8_t)(C.av_mallocz(C.size_t(extradataSize + C.AV_INPUT_BUFFER_PADDING_SIZE)))
-	if cp.c.extradata == nil {
-		return fmt.Errorf("failed to allocate extradata")
+	if cp.c.extradata = (*C.uint8_t)(C.av_mallocz(C.size_t(extradataSize + C.AV_INPUT_BUFFER_PADDING_SIZE))); cp.c.extradata == nil {
+		return fmt.Errorf("astiav: allocation is nil")
 	}
 
 	C.memcpy(unsafe.Pointer(cp.c.extradata), unsafe.Pointer(&extraData[0]), C.size_t(extradataSize))
--- a/codec_parameters.go
+++ b/codec_parameters.go
@@ -3,6 +3,10 @@ package astiav
 //#cgo pkg-config: libavcodec
 //#include <libavcodec/avcodec.h>
 import "C"
+import (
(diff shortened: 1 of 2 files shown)


codec_parameters.go
```diff
@@ -141,6 +145,27 @@ func (cp *CodecParameters) SetSampleAspectRatio(r Rational) {
...
+		return nil
+	}
+
+	if cp.c.extradata != nil {
+		C.av_freep(unsafe.Pointer(&cp.c.extradata))
+		cp.c.extradata_size = 0
+	}
+
+	extradataSize := len(extraData)
+	cp.c.extradata = (*C.uint8_t)(C.av_mallocz(C.size_t(extradataSize + C.AV_INPUT_BUFFER_PADDING_SIZE)))
```
asticode, line 159: Could you shorten with
```go
if cp.c.extradata = ...; cp.c.extradata == nil {
   ...
}
```

codec_parameters.go
```diff
@@ -141,6 +145,27 @@ func (cp *CodecParameters) SetSampleAspectRatio(r Rational) {
...
+	if cp.c.extradata == nil {
+		return fmt.Errorf("failed to allocate extradata")
```
asticode, line 161: Could you update the error message with `astiav: allocation is nil` instead ?

codec_parameters.go
```diff
@@ -141,6 +145,27 @@ func (cp *CodecParameters) SetSampleAspectRatio(r Rational) {
...
+	}
+
+	C.memcpy(unsafe.Pointer(cp.c.extradata), unsafe.Pointer(&extraData[0]), C.size_t(extradataSize))
+	cp.c.extradata_size = C.int(extradataSize)
+	return nil
+}
+
```
asticode, line 168: Could you add a `func (cp *CodecParameters) ExtraData() ([]byte, error)` method as well (using `bytesFromC` like [here](https://github.com/asticode/go-astiav/blob/master/packet.go#L89))?

codec_parameters_test.go
```diff
@@ -63,6 +63,9 @@ func TestCodecParameters(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 2, cp5.Channels())
 
+	err = cp5.SetExtraData([]byte{0, 0, 0, 1})
```
asticode, line 66: Could you use `cp6` (at the bottom of the test function) instead to test the new methods and simplify with `require.NoError(t, cp6.SetExtraData([]byte{0, 0, 0, 1}))`

codec_parameters_test.go
```diff
@@ -63,6 +63,9 @@ func TestCodecParameters(t *testing.T) {
...
+	require.NoError(t, err)
```
asticode, line 67: Could you test `cp6.Extradata()` as well?

asticode (commented):
Thanks for the PR!

Out of curiosity, would you mind sharing the extradata you needed to publish your rtmp stream? 🤔

Cacsjep:
done =)

Cacsjep:
When pushing a h264 rtmp stream you need to set the [H264 SPS/PPS](https://www.cardinalpeak.com/blog/the-h-264-sequence-parameter-set) as extra data, its like meta data, so that decoder knows how it can decode it correctly.  =)

asticode:
Thanks for the PR! ❤️ 

asticode:
> When pushing a h264 rtmp stream you need to set the [H264 SPS/PPS](https://www.cardinalpeak.com/blog/the-h-264-sequence-parameter-set) as extra data, its like meta data, so that decoder knows how it can decode it correctly. In my case a I got a h264 bytestream from a IP-Camera the first frame is an IDR Frame, this includes the SPS,PPS.

Ha you mean that you're using the `Packet.FromData()` method and therefore need to manually set the stream's codec parameters extradata whereas when using a FormatContext, it gets filled automatically?

Cacsjep:
> Ha you mean that you're using the `Packet.FromData()` method and therefore need to manually set the stream's codec parameters extradata whereas when using a FormatContext, it gets filled automatically?

correct =)

Cacsjep:
> Let me know whether you need a tag 👍
thanks no tag needed =)

PR #59, opened 2024-04-01
Add extradata methods also to cc

Cacsjep:


codec_context_test.go
```diff
@@ -132,6 +132,12 @@ func TestCodecContext(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 2, cp1.Channels())
 
+	cc6 := AllocCodecContext(nil)
+	require.NotNil(t, cc5)
```
asticode, line 136: This needs to be `cc6` and you need to add `defer cc6.Free()` below

asticode (commented):
I'd rather create an `extra_data.go` file (no test file needed) with 
```go
func getExtraData(d *C.uint8_t, size C.size_t) []byte {
  // Content of current CodecParameters.ExtraData()
}
func setExtraData(b []byte, d **C.uint8_t, size *C.size_t) error {
  // Content of current CodecParameters.SetExtraData()
} 
```
and use these 2 new functions in both `CodecContext` and `CodecParameters` 🤔

Cacsjep:
> I'd rather create an `extra_data.go` file (no test file needed) with
> 
> ```go
> func getExtraData(d *C.uint8_t, size C.size_t) []byte {
>   // Content of current CodecParameters.ExtraData()
> }
> func setExtraData(b []byte, d **C.uint8_t, size *C.size_t) error {
>   // Content of current CodecParameters.SetExtraData()
> } 
> ```
> 
> and use these 2 new functions in both `CodecContext` and `CodecParameters` 🤔

they share many similar fields like framerate, pixfmt etc.., then u have extra file just for these two things i dont know whats better.

asticode:
> they share many similar fields like framerate, pixfmt etc.., then u have extra file just for these two things i dont know whats better.

I think creating global unexported functions that take pointers to C fields like I suggest would be best:
```go
func getExtraData(d *C.uint8_t, size C.size_t) []byte {
  // Content of current CodecParameters.ExtraData()
}
func setExtraData(b []byte, d **C.uint8_t, size *C.size_t) error {
  // Content of current CodecParameters.SetExtraData()
} 
```
No need to create structs or anything else, just helper functions that unify the logic to retrieve and set the extradata. We may use them for other structs in the future.

For instance the content of `getExtraData` would be:
```go
	return bytesFromC(func(size *C.size_t) *C.uint8_t {
		*size = C.size_t(size)
		return d
	})
```

asticode:
@Cacsjep do you feel like implementing the requested changes? If not, that's ok, I'll have some time to implement them in the coming days 👍

Cacsjep:
> @Cacsjep do you feel like implementing the requested changes? If not, that's ok, I'll have some time to implement them in the coming days 👍

Oh, I'm sorry, I completely forgot, I don't have time at the moment. It would be cool if you make it, if not I can do it next month or so =)

asticode:
No worries, fyi I've merged in `master` the extra data methods to `CodecContext` 👍

PR #61, opened 2024-05-13
Add bit stream filter support

leandromoreira:
Add support to [Bit Stream Filter](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html).

> Bitstream filters transform encoded media data without decoding it. This allows e.g. manipulating various header values. Bitstream filters operate on [AVPackets](https://ffmpeg.org/doxygen/5.1/structAVPacket.html).
>
> The bitstream filtering API is centered around two structures: [AVBitStreamFilter](https://ffmpeg.org/doxygen/5.1/structAVBitStreamFilter.html) and [AVBSFContext](https://ffmpeg.org/doxygen/5.1/structAVBSFContext.html). The former represents a bitstream filter in abstract, the latter a specific filtering process. Obtain an [AVBitStreamFilter](https://ffmpeg.org/doxygen/5.1/structAVBitStreamFilter.html) using [av_bsf_get_by_name()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#gae491493190b45698ebd43db28c4e8fe9) or [av_bsf_iterate()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#gaff3224f9026f829d25557cf85c9e9e12), then pass it to [av_bsf_alloc()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#ga7da65af303e20c9546e15ec266b182c1) to create an [AVBSFContext](https://ffmpeg.org/doxygen/5.1/structAVBSFContext.html). [Fill](https://ffmpeg.org/doxygen/5.1/structFill.html) in the user-settable [AVBSFContext](https://ffmpeg.org/doxygen/5.1/structAVBSFContext.html) fields, as described in its documentation, then call [av_bsf_init()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#ga242529d54013acf87e94273d298a5ff2) to prepare the filter context for use.
>
> Submit packets for filtering using [av_bsf_send_packet()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#gaada9ea8f08d3dcf23c14564dbc88992c), obtain filtered results with [av_bsf_receive_packet()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#ga7fffb6c87b91250956e7a2367af56b38). When no more input packets will be sent, submit a NULL [AVPacket](https://ffmpeg.org/doxygen/5.1/structAVPacket.html) to signal the end of the stream to the filter. [av_bsf_receive_packet()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#ga7fffb6c87b91250956e7a2367af56b38) will then return trailing packets, if any are produced by the filter.
>
> Finally, free the filter context with [av_bsf_free()](https://ffmpeg.org/doxygen/5.1/group__lavc__bsf.html#ga08d53431e76355f88e27763b1940df4f).

src: https://github.com/FFmpeg/FFmpeg/blob/release/5.1/libavcodec/bsf.h

### DIFF
--- a/astiav_test.go
+++ b/astiav_test.go
@@ -12,16 +12,6 @@ import (
 
 var globalHelper = newHelper()
 
-type testConfig struct {
-	withBitStreamFilter *string
-}
-
-func withBitStreamFilter(fn string) func(c *testConfig) {
-	return func(c *testConfig) {
-		c.withBitStreamFilter = &fn
-	}
-}
-
 func TestMain(m *testing.M) {
 	// Make sure to exit with the proper code
 	var code int
@@ -95,19 +85,22 @@ func (h *helper) inputFormatContext(name string) (fc *FormatContext, err error) 
 	return
 }
 
-func (h *helper) bitStreamFilterContext(fc *FormatContext, pkt *Packet, tc *testConfig) (*BitStreamFilterContext, error) {
+func (h *helper) bitStreamFilterContext(fc *FormatContext, si int, fn string) (*BitStreamFilterContext, error) {
 	var cs *Stream
 	for _, s := range fc.Streams() {
-		if s.Index() != pkt.StreamIndex() {
+		if s.Index() != si {
 			continue
 		}
 		cs = s
 		break
 	}
+	if cs == nil {
+		return nil, fmt.Errorf("astiav_test: could not find a stream to the packet")
+	}
 
-	bsf := FindBitStreamFilterByName(*tc.withBitStreamFilter)
+	bsf := FindBitStreamFilterByName(fn)
 	if bsf == nil {
-		return nil, fmt.Errorf("astiav_test: cannot find the bit stream filter %s", *tc.withBitStreamFilter)
+		return nil, fmt.Errorf("astiav_test: cannot find the bit stream filter %s", fn)
 	}
 
 	var bsfc *BitStreamFilterContext
@@ -129,12 +122,7 @@ func (h *helper) bitStreamFilterContext(fc *FormatContext, pkt *Packet, tc *test
 	return bsfc, nil
 }
 
-func (h *helper) inputFirstPacket(name string, options ...func(*testConfig)) (pkt *Packet, err error) {
-	testConfig := &testConfig{}
-	for _, o := range options {
-		o(testConfig)
-	}
-
+func (h *helper) inputFirstPacket(name string) (pkt *Packet, err error) {
 	h.m.Lock()
 	i, ok := h.inputs[name]
 	if ok && i.firstPkt != nil {
@@ -161,30 +149,49 @@ func (h *helper) inputFirstPacket(name string, options ...func(*testConfig)) (pk
 		return
 	}
 
-	if testConfig.withBitStreamFilter != nil {
-		var bsfc *BitStreamFilterContext
-
-		bsfc, err = h.bitStreamFilterContext(fc, pkt, testConfig)
-		if err != nil {
-			pkt = nil
-			return
-		}
-
-		if err = bsfc.SendPacket(pkt); err != nil {
-			err = fmt.Errorf("astiav_test: error while sending the packet %w", err)
-			pkt = nil
-			return
-		}
-		if err = bsfc.ReceivePacket(pkt); err != nil {
-			pkt = nil
-			err = fmt.Errorf("astiav_test: error while receiving the packet %w", err)
-			return
-		}
-	}
-
 	h.m.Lock()
 	h.inputs[name].firstPkt = pkt
 	h.m.Unlock()
+	return
+}
+
+func (h *helper) inputFirstPacketWithBitStreamFilter(name string, fn string) (pkt *Packet, err error) {
+	var fc *FormatContext
+	if fc, err = h.inputFormatContext(name); err != nil {
+		err = fmt.Errorf("astiav_test: getting input format context failed")
+		return
+	}
+
+	pkt = AllocPacket()
+	if pkt == nil {
+		err = errors.New("astiav_test: pkt is nil")
+		return
+	}
+	h.closer.Add(pkt.Free)
+
+	if err = fc.ReadFrame(pkt); err != nil {
+		err = fmt.Errorf("astiav_test: reading frame failed: %w", err)
+		return
+	}
+
+	var bsfc *BitStreamFilterContext
+
+	bsfc, err = h.bitStreamFilterContext(fc, pkt.StreamIndex(), fn)
+	if err != nil {
+		pkt = nil
+		return
+	}
+
+	if err = bsfc.SendPacket(pkt); err != nil {
+		err = fmt.Errorf("astiav_test: error while sending the packet %w", err)
+		pkt = nil
+		return
+	}
+	if err = bsfc.ReceivePacket(pkt); err != nil {
+		pkt = nil
+		err = fmt.Errorf("astiav_test: error while receiving the packet %w", err)
+		return
+	}
 	return
 }
 
--- a/astiav_test.go
+++ b/astiav_test.go
@@ -155,7 +155,7 @@ func (h *helper) inputFirstPacket(name string) (pkt *Packet, err error) {
 	return
 }
 
-func (h *helper) inputFirstPacketWithBitStreamFilter(name string, fn string) (pkt *Packet, err error) {
+func (h *helper) inputFirstVideoPacketWithBitStreamFilter(name string, fn string) (pkt *Packet, err error) {
 	var fc *FormatContext
 	if fc, err = h.inputFormatContext(name); err != nil {
 		err = fmt.Errorf("astiav_test: getting input format context failed")
@@ -168,14 +168,32 @@ func (h *helper) inputFirstPacketWithBitStreamFilter(name string, fn string) (pk
 		return
 	}
 	h.closer.Add(pkt.Free)
+	var foundVideo bool
 
-	if err = fc.ReadFrame(pkt); err != nil {
-		err = fmt.Errorf("astiav_test: reading frame failed: %w", err)
+	for {
+		if err = fc.ReadFrame(pkt); err != nil {
+			err = fmt.Errorf("astiav_test: reading frame failed: %w", err)
+			return
+		}
+
+		for _, s := range fc.Streams() {
+			if s.Index() == pkt.StreamIndex() && s.CodecParameters().CodecID() == CodecIDH264 {
+				foundVideo = true
+				break
+			}
+		}
+		if foundVideo {
+			break
+		}
+	}
+
+	if !foundVideo {
+		err = fmt.Errorf("astiav_test: there must be an h264 stream")
+		pkt = nil
 		return
 	}
 
 	var bsfc *BitStreamFilterContext
-
 	bsfc, err = h.bitStreamFilterContext(fc, pkt.StreamIndex(), fn)
 	if err != nil {
 		pkt = nil
--- a/astiav_test.go
+++ b/astiav_test.go
@@ -12,6 +12,16 @@ import (
 
 var globalHelper = newHelper()
 
+type testConfig struct {
+	withBitStreamFilter *string
+}
+
+func withBitStreamFilter(fn string) func(c *testConfig) {
+	return func(c *testConfig) {
+		c.withBitStreamFilter = &fn
+	}
+}
+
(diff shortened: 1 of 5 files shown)


bit_stream_filter.go
```diff
@@ -0,0 +1,28 @@
...
+// https://github.com/FFmpeg/FFmpeg/blob/release/5.1/libavcodec/bsf.h#L111
+type BitStreamFilter struct {
+	c *C.struct_AVBitStreamFilter
+}
+
+func newBitStreamFilterFromC(c *C.struct_AVBitStreamFilter) *BitStreamFilter {
+	if c == nil {
+		return nil
+	}
+	cc := &BitStreamFilter{c: c}
```
asticode, line 20: Could you `return &BitStreamFilter{c: c}` right away without creating the intermediary `cc`?

bit_stream_filter_context.go
```diff
@@ -0,0 +1,81 @@
...
+//#cgo pkg-config: libavcodec
+//#include <libavcodec/bsf.h>
+import "C"
+import (
+	"errors"
+	"unsafe"
+)
+
+// https://github.com/FFmpeg/FFmpeg/blob/release/5.1/libavcodec/bsf.h#L68
+type BSFContext struct {
```
asticode, line 12: Could you rename the struct to `BitStreamFilterContext`?

bit_stream_filter_context.go
```diff
@@ -0,0 +1,81 @@
...
+	c *C.struct_AVBSFContext
+}
+
+func newBSFContextFromC(c *C.struct_AVBSFContext) *BSFContext {
+	if c == nil {
+		return nil
+	}
+	bsfCtx := &BSFContext{c: c}
```
asticode, line 20: Could you rename all `bsfCtx` occurences in this file to `bsfc` instead?

bit_stream_filter_context.go
```diff
@@ -0,0 +1,81 @@
...
+	classers.set(bsfCtx)
+	return bsfCtx
+}
+
+var _ Classer = (*BSFContext)(nil)
+
+func AllocBitStreamContext(f *BitStreamFilter) (*BSFContext, error) {
```
asticode, line 27: Could you rename this function to `AllocBitStreamFilterContext` instead?

bit_stream_filter_context.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	return newBSFContextFromC(bsfCtx), nil
+}
+
+func (bsfCtx *BSFContext) Class() *Class {
+	return newClassFromC(unsafe.Pointer(bsfCtx.c))
+}
+
+func (bsfCtx *BSFContext) Init() error {
```
asticode, line 44: Could you rename this function to `Initialize`?

bit_stream_filter_context.go
```diff
@@ -0,0 +1,81 @@
...
+	return newError(C.av_bsf_init(bsfCtx.c))
+}
+
+func (bsfCtx *BSFContext) SendPacket(p *Packet) error {
+	if p == nil {
```
asticode, line 49: The doc states that the packet can be nil (you need this to flush the context properly). Therefore I would do something like this instead:
```go
var pc *C.struct_AVPacket
if p != nil {
  pc = p.c
}
return newError(C.av_bsf_send_packet(bsfc.c, pc))
```

asticode (commented):
Thanks for the PR ❤️ 

I've made some comments 👍 

Also, could you add tests?
- for the bit stream filter, the best would be to add a `Name() string` function in the struct, and in the test find a filter available by default and assert its name.
- for the bit stream filter context, the best would be to test allocating a context, initializing it and freeing it. Before freeing it, that would be a good idea to test the getters and setters as well (for the Class you can use something similar to [this](https://github.com/asticode/go-astiav/blob/master/format_context_test.go#L30)). If you find a way to test Send/ReceivePacket that would be even better, but that's optionnal.

leandromoreira (commented):
done

asticode:
Nice! Did you see the part of my comment mentioning tests as well?

leandromoreira:
> Thanks for the PR ❤️
> 
> I've made some comments 👍
> 
> Also, could you add tests?
> 
>     * for the bit stream filter, the best would be to add a `Name() string` function (and a `String() string` function that would return the result of .Name()) in the struct, and in the test find a filter available by default and assert its name.
> 
>     * for the bit stream filter context, the best would be to test allocating a context, initializing it and freeing it. Before freeing it, that would be a good idea to test the getters and setters as well (for the Class you can use something similar to [this](https://github.com/asticode/go-astiav/blob/master/format_context_test.go#L30)). If you find a way to test Send/ReceivePacket that would be even better, but that's optionnal.

Will add tests later :)

leandromoreira:
The tests I introduced are passing:

```
 === RUN   TestBitStreamFilterContext
--- PASS: TestBitStreamFilterContext (0.00s)
=== RUN   TestBitStreamFilter
--- PASS: TestBitStreamFilter (0.00s)
```
They shouldn't cause any side effect though.

bit_stream_filter_test.go
```diff
@@ -0,0 +1,21 @@
+package astiav
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestBitStreamFilter(t *testing.T) {
+	bitStreamFilterNames := []string{"h264_mp4toannexb", "null", "setts"}
```
asticode, line 10: I don't think we need to test 3 filters here, could you only test the `null` filter?
Testing for an invalid filter name is a good idea however 👍

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestBitStreamFilterContext(t *testing.T) {
+	bsf := FindBitStreamFilterByName("null")
+	require.NotNil(t, bsf)
+
+	bsfc, err := AllocBitStreamFilterContext(bsf)
+	require.NotNil(t, bsfc)
+	require.Nil(t, err)
```
asticode, line 15: Could you use `require.NoError` instead?

astiav_test.go
```diff
@@ -85,6 +85,43 @@ func (h *helper) inputFormatContext(name string) (fc *FormatContext, err error)
 	return
 }
 
+func (h *helper) bitStreamFilterContext(fc *FormatContext, si int, fn string) (*BitStreamFilterContext, error) {
```
asticode, line 88: I don't feel the way we test `SendPacket` and `ReceivePacket` is sufficient right now but I'm not sure how to do it differently. In the meantime I'd rather not add it in the PR and add a TODO instead.
Therefore could you remove the changes in `astiav_test.go` and in `bit_stream_filter_context_test.go` remove all things needed to test both `SendPacket` and `ReceivePacket`?

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,42 @@
...
+	bsfc.SetTimeBaseIn(NewRational(15, 1))
+	require.Equal(t, NewRational(15, 1), bsfc.TimeBaseIn())
+
+	fc, err := globalHelper.inputFormatContext("video.mp4")
+	require.NoError(t, err)
+	ss := fc.Streams()
+	require.Len(t, ss, 2)
+	s1 := ss[0]
+
+	cp1 := s1.CodecParameters()
```
asticode, line 31: Instead of using `s1.CodecParameters()` could you alloc a new codec parameter (don't forget to free it) whom you'll can only set the codec id in order to test `CodecParameterIn` setter and getter?

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,42 @@
...
+	defer bsfc.Free()
+
+	cl := bsfc.Class()
+	require.NotNil(t, cl)
+	require.Equal(t, "AVBSFContext", cl.Name())
+
+	bsfc.SetTimeBaseIn(NewRational(15, 1))
+	require.Equal(t, NewRational(15, 1), bsfc.TimeBaseIn())
+
+	fc, err := globalHelper.inputFormatContext("video.mp4")
```
asticode, line 25: Following my previous comment, this would need to be removed

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,42 @@
...
+	require.NoError(t, err)
+	ss := fc.Streams()
+	require.Len(t, ss, 2)
+	s1 := ss[0]
+
+	cp1 := s1.CodecParameters()
+	bsfc.SetCodecParametersIn(cp1)
+	require.Equal(t, int64(441324), bsfc.CodecParametersIn().BitRate())
+
+	// video.mp4 bit stream h264 format is avcc
```
asticode, line 35: Following my previous comment, this would need to be removed

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	fc, err := globalHelper.inputFormatContext("video.mp4")
+	require.NoError(t, err)
+	ss := fc.Streams()
+	require.Len(t, ss, 2)
+	s1 := ss[0]
+
+	cp1 := s1.CodecParameters()
+	bsfc.SetCodecParametersIn(cp1)
+	require.Equal(t, int64(441324), bsfc.CodecParametersIn().BitRate())
```
asticode, line 33: You'll need to test the codec id instead

asticode (changes requested):

asticode:
> The tests I introduced are passing:
> 
> ```
>  === RUN   TestBitStreamFilterContext
> --- PASS: TestBitStreamFilterContext (0.00s)
> === RUN   TestBitStreamFilter
> --- PASS: TestBitStreamFilter (0.00s)
> ```
> 
> They shouldn't cause any side effect though.

Yeah testing `SendPacket` and `ReceivePacket` like this doesn't feel right, but I have no alternative for now. Could add `// TODO Test SendPacket and ReceivePacket` at the end of `TestBitStreamFilterContext` in `bit_stream_filter_context_test.go` though?

asticode:
Windows tests seem to be failing (without any error message, that would be too easy otherwise 🤦 ), I'll take a look

leandromoreira:
> Windows tests seem to be failing (without any error message, that would be too easy otherwise 🤦 ), I'll take a look

:( I don't have a windows to test, otherwise I'd try to help.

BTW, is it easy to run tests locally on macos?

asticode:
> :( I don't have a windows to test, otherwise I'd try to help.

I'm using Github actions to run tests on windows and linux

> BTW, is it easy to run tests locally on macos?

Sure, as long as ffmpeg is installed and the proper environment variables are setup, a simple `go test` does the trick

bit_stream_filter_context_test.go
```diff
@@ -0,0 +1,34 @@
...
+	cl := bsfc.Class()
+	require.NotNil(t, cl)
+	require.Equal(t, "AVBSFContext", cl.Name())
+
+	bsfc.SetTimeBaseIn(NewRational(15, 1))
+	require.Equal(t, NewRational(15, 1), bsfc.TimeBaseIn())
+
+	cp1 := AllocCodecParameters()
+	require.NotNil(t, cp1)
+	defer cp1.Free()
```
asticode, line 27: OK so I figured out why tests were failing on Windows, basically you need to remove the `defer cp1.Free()` since the context seems to be taking ownership of the codec parameters and free it when calling `bsfc.Free()`

asticode (changes requested):

bit_stream_filter_context_test.go
leandromoreira: thanks, great catch! :)

leandromoreira: probably calling free twice on unix didn't raise an error while on windows it'll

CypherpunkSamurai:
> > :( I don't have a windows to test, otherwise I'd try to help.
> 
> I'm using Github actions to run tests on windows and linux
> 
> > BTW, is it easy to run tests locally on macos?
> 
> Sure, as long as ffmpeg is installed and the proper environment variables are setup, a simple `go test` does the trick

Can you please help me, I have the ffmpeg binary available in path, yet when i run go build it seems to fail.

asticode:
You need to install `pkg-config` on the machine and it should be available in your `$PATH` 👍 

Let me know whether you need more help

Issue #66, opened 2024-08-22
Question: How can i set metadata in output FormatContext?


Mr3h1v:
Add `outputFormatCtx.Metadata().Set("key", "value", 0)` after AllocOutputFormatContext, it will panic because Metadata() return nil.

when i add `outputFormatCtx.Metadata().Set("key", "value", 0)` after WriteHeader, it does not work.

asticode:
Fixed by #67 👍

PR #67, opened 2024-08-22
Add SetMetadata for FormatContext

Mr3h1v:
#66 
support set metadata

### DIFF
--- a/format_context.go
+++ b/format_context.go
@@ -106,6 +106,14 @@ func (fc *FormatContext) Metadata() *Dictionary {
 	return newDictionaryFromC(fc.c.metadata)
 }
 
+func (fc *FormatContext) SetMetadata(d *Dictionary) {
+	if d == nil {
+		fc.c.metadata = nil
+	} else {
+		fc.c.metadata = d.c
+	}
+}
+
 func (fc *FormatContext) NbStreams() int {
 	return int(fc.c.nb_streams)
 }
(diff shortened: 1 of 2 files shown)


format_context.go
```diff
@@ -106,6 +106,13 @@ func (fc *FormatContext) Metadata() *Dictionary {
 	return newDictionaryFromC(fc.c.metadata)
 }
 
+func (fc *FormatContext) Metadata(d *Dictionary) {
```
asticode, line 109: Shouldn't it be `SetMetadata` instead?

format_context.go
```diff
@@ -106,6 +106,13 @@ func (fc *FormatContext) Metadata() *Dictionary {
...
+	if d == nil {
```
asticode, line 110: I'd rather not ignore if `d == nil` in case someone would want to reset the metadata to `nil`

asticode (changes requested):
Thanks for the PR 👍 

Could you also add a simple test [here](https://github.com/asticode/go-astiav/blob/master/format_context_test.go#L51), something like:
```go
d := NewDictionnary()
d.Set("k", "v", ...)
fc3.SetMetadata(d)
require.Equal(t, "v", fc3.Metadata().Get("k")
require.Equal(t, d
```

asticode (approved):

asticode:
Thanks again for the PR ❤️ 

Let me know whether you need me to make a tag 👍

Mr3h1v:
> Thanks again for the PR ❤️
> 
> Let me know whether you need me to make a tag 👍

I do need a tag. please make one

asticode:
I've created a `v0.18.0` tag 👍

PR #70, opened 2024-09-04
Add hardware frame context for hardware video encoding

l0rem1psum:
With the addition of hardware frame context, hardware video encoding can be achieved. The official vaapi_encode.c example can now be implemented in Go. I personally tried with CUDA and it works similar to vaapi.

This PR does not have UTs yet. Please help review the changes before I can proceed to write some UTs.

### DIFF
--- a/codec_context.go
+++ b/codec_context.go
@@ -37,6 +37,7 @@ type CodecContext struct {
 	c *C.struct_AVCodecContext
 	// We need to store this to unref it properly
 	hdc *HardwareDeviceContext
+	hfc *HardwareFrameContext
 }
 
 func newCodecContextFromC(c *C.struct_AVCodecContext) *CodecContext {
@@ -62,6 +63,10 @@ func (cc *CodecContext) Free() {
 	if cc.hdc != nil {
 		C.av_buffer_unref(&cc.hdc.c)
 		cc.hdc = nil
+	}
+	if cc.hfc != nil {
+		C.av_buffer_unref(&cc.hfc.c)
+		cc.hfc = nil
 	}
 	classers.del(cc)
 	C.avcodec_free_context(&cc.c)
@@ -341,6 +346,16 @@ func (cc *CodecContext) SetHardwareDeviceContext(hdc *HardwareDeviceContext) {
 	}
 }
 
+func (cc *CodecContext) SetHardwareFrameContext(hfc *HardwareFrameContext) {
+	if cc.c.hw_frames_ctx != nil {
+		C.av_buffer_unref(&cc.c.hw_frames_ctx)
+	}
+	cc.hfc = hfc
+	if cc.hfc != nil {
+		cc.c.hw_frames_ctx = C.av_buffer_ref(cc.hfc.c)
+	}
+}
+
 func (cc *CodecContext) ExtraHardwareFrames() int {
 	return int(cc.c.extra_hw_frames)
 }
@@ -391,5 +406,4 @@ func goAstiavCodecContextGetFormat(cc *C.struct_AVCodecContext, pfsCPtr *C.enum_
 
 	// Callback
 	return C.enum_AVPixelFormat(c(pfs))
-
 }
--- a/frame.go
+++ b/frame.go
@@ -33,6 +33,10 @@ func (f *Frame) AllocBuffer(align int) error {
 	return newError(C.av_frame_get_buffer(f.c, C.int(align)))
 }
 
+func (f *Frame) AllocHWBuffer(hfc *HardwareFrameContext, align int) error {
+	return newError(C.av_hwframe_get_buffer(hfc.c, f.c, C.int(align)))
+}
+
 func (f *Frame) AllocImage(align int) error {
 	return newError(C.av_image_alloc(&f.c.data[0], &f.c.linesize[0], f.c.width, f.c.height, (C.enum_AVPixelFormat)(f.c.format), C.int(align)))
 }
--- a/hardware_frame_context.go
+++ b/hardware_frame_context.go
@@ -1,0 +1,48 @@
+package astiav
+
+//#include <libavcodec/avcodec.h>
+import "C"
+import (
+	"fmt"
+	"unsafe"
+)
+
+type HardwareFrameContext struct {
+	c *C.struct_AVBufferRef
+}
+
+func AllocHardwareFrameContext(hdc *HardwareDeviceContext) (*HardwareFrameContext, error) {
+	if hfc := C.av_hwframe_ctx_alloc(hdc.c); hfc != nil {
+		return &HardwareFrameContext{c: hfc}, nil
+	}
+	return nil, fmt.Errorf("failed to allocate hardware frame context")
+}
+
+func (hfc *HardwareFrameContext) SetWidth(width int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.width = C.int(width)
+}
+
+func (hfc *HardwareFrameContext) SetHeight(height int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.height = C.int(height)
+}
+
+func (hfc *HardwareFrameContext) SetFormat(format PixelFormat) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.format = C.enum_AVPixelFormat(format)
+}
+
+func (hfc *HardwareFrameContext) SetSWFormat(swFormat PixelFormat) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.sw_format = C.enum_AVPixelFormat(swFormat)
+}
+
+func (hfc *HardwareFrameContext) SetInitialPoolSize(initialPoolSize int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.initial_pool_size = C.int(initialPoolSize)
+}
+
+func (hfc *HardwareFrameContext) Init() error {
+	return newError(C.av_hwframe_ctx_init(hfc.c))
+}


codec_context.go
```diff
@@ -341,6 +346,16 @@ func (cc *CodecContext) SetHardwareDeviceContext(hdc *HardwareDeviceContext) {
 	}
 }
 
+func (cc *CodecContext) SetHardwareFrameContext(hfc *HardwareFrameContext) {
+	if cc.c.hw_frames_ctx != nil {
```
asticode, line 350: Can you use `cc.hfc.c` instead of `cc.c.hw_frames_ctx` here and the line below as well?

frame.go
```diff
@@ -33,6 +33,10 @@ func (f *Frame) AllocBuffer(align int) error {
 	return newError(C.av_frame_get_buffer(f.c, C.int(align)))
 }
 
+func (f *Frame) AllocHWBuffer(hfc *HardwareFrameContext, align int) error {
```
asticode, line 36: I don't think you need `align int` here since the doc states that the final argument of `av_hwframe_get_buffer` is `flags` and is currently unused and should be set to 0

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"unsafe"
+)
+
+type HardwareFrameContext struct {
+	c *C.struct_AVBufferRef
+}
+
+func AllocHardwareFrameContext(hdc *HardwareDeviceContext) (*HardwareFrameContext, error) {
+	if hfc := C.av_hwframe_ctx_alloc(hdc.c); hfc != nil {
```
asticode, line 15: Can you create a `newHardwareFrameContextFromC` func similar to [this](https://github.com/asticode/go-astiav/blob/master/output_format.go#L13) and use it here instead of `&HardwareFrameContext{c: hfc}`?

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"fmt"
+	"unsafe"
+)
+
+type HardwareFrameContext struct {
+	c *C.struct_AVBufferRef
+}
+
+func AllocHardwareFrameContext(hdc *HardwareDeviceContext) (*HardwareFrameContext, error) {
```
asticode, line 14: Can you not return an error here but only a `*HardwareFrameContext` and let the developer checks whether it's nil or not?

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+	if hfc := C.av_hwframe_ctx_alloc(hdc.c); hfc != nil {
+		return &HardwareFrameContext{c: hfc}, nil
+	}
+	return nil, fmt.Errorf("failed to allocate hardware frame context")
+}
+
+func (hfc *HardwareFrameContext) SetWidth(width int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
```
asticode, line 22: Can you create a `func (hfc *HardwareFrameContext) data() *C.AVHWFramesContext` and use it all below functions instead of repeating `frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))`?

It would end up being `hfc.data().width = C.int(width)` for instance

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+	frameCtx.width = C.int(width)
+}
+
+func (hfc *HardwareFrameContext) SetHeight(height int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.height = C.int(height)
+}
+
+func (hfc *HardwareFrameContext) SetFormat(format PixelFormat) {
```
asticode, line 31: Can you rename this function to `SetPixelFormat`?

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.format = C.enum_AVPixelFormat(format)
+}
+
+func (hfc *HardwareFrameContext) SetSWFormat(swFormat PixelFormat) {
```
asticode, line 36: Can you rename this function to `SetSoftwarePixelFormat`?

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.sw_format = C.enum_AVPixelFormat(swFormat)
+}
+
+func (hfc *HardwareFrameContext) SetInitialPoolSize(initialPoolSize int) {
+	frameCtx := (*C.AVHWFramesContext)(unsafe.Pointer((hfc.c.data)))
+	frameCtx.initial_pool_size = C.int(initialPoolSize)
+}
+
+func (hfc *HardwareFrameContext) Init() error {
```
asticode, line 46: Can you rename this function to `Initialize`?

frame.go
asticode, line 36: Can you rename this function to `AllocHardwareBuffer`?

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
+package astiav
+
+//#include <libavcodec/avcodec.h>
```
asticode, line 3: This is weird to me that you don't need 
```go
//#cgo pkg-config: libavutil libavcodec
//#include <libavcodec/avcodec.h>
//#include <libavutil/hwcontext.h>
```
here 🤔

hardware_frame_context.go
```diff
@@ -0,0 +1,48 @@
...
+import "C"
+import (
+	"fmt"
+	"unsafe"
+)
+
+type HardwareFrameContext struct {
```
asticode, line 10: Can you add this comment above `// https://github.com/FFmpeg/FFmpeg/blob/n7.0/libavutil/hwcontext.h#L115`?

asticode (changes requested):
Thanks for this PR!

I don't know if unit tests will be possible since for instance Github actions don't have GPUs, but that's possible, that's indeed something that would need to be added.

Also, since the `vaapi_encode.c` would be possible, we would need to add it in the list of examples as well 👍

l0rem1psum:
Thanks for such a detailed review! I will make the relevant changes once I get back.

asticode:
> Thanks for such a detailed review! I will make the relevant changes once I get back.

No problem. Don't forget to add a comment here once you're done as I won't get notified by Github otherwise 👍

hardware_frame_context.go
l0rem1psum: I believe `#cgo` directives (including `#cgo CPPFLAGS`, `#cgo LDFLAGS`, etc) only need to specified once per package, whereas `#include` needs to be specified every file where it's needed.

Not sure about what is the best practice tho, but my preference both for my projects and projects at work is to add a `cgo.go` file to include all relevant cgo flags for a more centralized and easier management.

l0rem1psum:
@asticode All above comments were fixed. Please review when you have time.

l0rem1psum:
Is the `vaapi_encode.c` Go example required for this pr to get merged? Or should we do it in a separate one?

hardware_frame_context.go
asticode: Understood. Can you still add the 3 lines? I'll take some time afterward to clean things up 👍

asticode (changes requested):
I'd rather have the example in this PR so that everything is in the same place 👍 When implementing the example try to stick as close as possible to the C example but also take inspiration in other Go examples in this package and feel free to get away from the C example if that makes things easier to understand 👍

hardware_frame_context.go
asticode: In fact can you only add the 2 `include` directives?

asticode:
Did you have the chance to finalize this PR by any chance?

l0rem1psum:
> Did you have the chance to finalize this PR by any chance?

I will have to find some time next week for this.

l0rem1psum:
The hardware video encoding example is now added. 

The `vaapi_encode.c` example in the ffmpeg repo has only hardcoded support for vaapi. I do not have a device that supports vaapi. Or maybe I do, but I'm not familiar with the ecosystem. Therefore, I took the liberty to remove the hardcoded parts to make the example support other hardware devices like `cuda`. I think this example will be more helpful for most people.

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		log.Fatal(fmt.Errorf("main: initializing hardware frame context failed: %w", err))
+	}
+	encCodecContext.SetHardwareFrameContext(hardwareFrameCtx)
+
+	// Open codec context
+	if err := encCodecContext.Open(encCodec, nil); err != nil {
+		log.Fatal(fmt.Errorf("main: opening codec context failed: %w", err))
+	}
+
+	frameIndex := 0
```
asticode, line 116: Can you remove this variable?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// Draw frames, upload them to hardware devices, and encode them
```
asticode, line 118: Can you replace this comment with `// Loop`?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+	for {
+		// Alloc software frame
```
asticode, line 120: To ease frames freeing, I'd rather use a closure here therefore could you wrap the logic inside the `for` loop with something like:
```go
// We use a closure to ease frame freeing
func() {
  // Alloc software frame
  softwareFrame := astiav.AllocFrame()
  defer softwareFrame.Free()

  // Alloc hardware frame
  hardwareFrame := astiav.AllocFrame()
  defer hardwareFrame.Free()

  ... the rest of the logic
}()
```

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		softwareFrame.SetWidth(*width)
+		softwareFrame.SetHeight(*height)
+		softwareFrame.SetPixelFormat(astiav.PixelFormatNv12)
+
+		// Alloc software frame buffer
+		if err := softwareFrame.AllocBuffer(0); err != nil {
+			log.Fatal(fmt.Errorf("main: allocating buffer failed: %w", err))
+		}
+
+		// Fill software frame
```
asticode, line 133: Although your idea is really cool, I'd rather keep things simple regarding the source frame and only do
```go
// Fill software frame with black
if err = srcFrame.ImageFillBlack(); err != nil {
	log.Fatal(fmt.Errorf("main: filling source frame with black image failed: %w", err))
}
```

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+			packet.Unref()
+		}
+		packet.Free()
+
+		frameIndex++
+		log.Printf("Finished encoding frame %d\n", frameIndex)
+	}
+}
+
+func MakeNV12MovingCheckerboardPattern(width, height, blockSize, frame int) ([]byte, []byte) {
```
asticode, line 179: Can you remove this?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		hardwareFrame := astiav.AllocFrame()
+
+		// Alloc hardware frame buffer
+		if err := hardwareFrame.AllocHardwareBuffer(hardwareFrameCtx); err != nil {
+			log.Fatal(fmt.Errorf("main: allocating hardware buffer failed: %w", err))
+		}
+
+		// Upload software frame to hardware frame
+		if err := softwareFrame.TransferHardwareData(hardwareFrame); err != nil {
+			log.Fatal(fmt.Errorf("main: uploading from frame failed: %w", err))
```
asticode, line 148: Can you replace the error message to `transferring hardware data failed: %w`

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+			log.Fatal(fmt.Errorf("main: allocating buffer failed: %w", err))
+		}
+
+		// Fill software frame
+		yPlane, uvPlane := MakeNV12MovingCheckerboardPattern(*width, *height, *patternGridSize, frameIndex)
+		softwareFrame.SetData(0, yPlane)
+		softwareFrame.SetData(1, uvPlane)
+
+		// Alloc hardware frame
+		hardwareFrame := astiav.AllocFrame()
```
asticode, line 139: Moved above, you can remove here

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		// Alloc hardware frame buffer
+		if err := hardwareFrame.AllocHardwareBuffer(hardwareFrameCtx); err != nil {
+			log.Fatal(fmt.Errorf("main: allocating hardware buffer failed: %w", err))
+		}
+
+		// Upload software frame to hardware frame
+		if err := softwareFrame.TransferHardwareData(hardwareFrame); err != nil {
+			log.Fatal(fmt.Errorf("main: uploading from frame failed: %w", err))
+		}
+		softwareFrame.Free()
```
asticode, line 150: Moved above, you can remove here

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+
+		// Encode frame
+		if err := encCodecContext.SendFrame(hardwareFrame); err != nil {
+			log.Fatal(fmt.Errorf("main: sending frame failed: %w", err))
+		}
+		hardwareFrame.Free()
```
asticode, line 156: Moved above, you can remove here

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+
+		// Receive packet
+		packet := astiav.AllocPacket()
```
asticode, line 159: Can you 
```go
// Alloc packet
packet := astiav.AllocPacket()
defer packet.Free()
```
once before entering the first `for` loop?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		softwareFrame.Free()
+
+		// Encode frame
+		if err := encCodecContext.SendFrame(hardwareFrame); err != nil {
+			log.Fatal(fmt.Errorf("main: sending frame failed: %w", err))
+		}
+		hardwareFrame.Free()
+
+		// Receive packet
```
asticode, line 158: Can you replace this comment with `// Loop`?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		packet := astiav.AllocPacket()
+		for {
+			if err := encCodecContext.ReceivePacket(packet); err != nil {
```
asticode, line 161: Can you replace this with
```go
// Receive packet
if err = encCodecContext.ReceivePacket(packet); err != nil {
	if errors.Is(err, astiav.ErrEof) || errors.Is(err, astiav.ErrEagain) {
		err = nil
		break
	}
	err = fmt.Errorf("main: receiving packet failed: %w", err)
	return
}
```

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+				break
+			}
+
+			// Write packet
```
asticode, line 165: You can remove this

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+			if _, err := output.Write(packet.Data()); err != nil {
+				log.Fatal(fmt.Errorf("main: writing packet failed: %w", err))
+			}
+
+			packet.Unref()
```
asticode, line 170: Can you remove this? you shouldn't need it

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		}
+		packet.Free()
```
asticode, line 172: Move above, you can remove here

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+
+		frameIndex++
```
asticode, line 174: Can you remove this?

examples/hardware_encoding/main.go
```diff
@@ -0,0 +1,204 @@
...
+		for {
+			if err := encCodecContext.ReceivePacket(packet); err != nil {
+				break
+			}
+
+			// Write packet
+			if _, err := output.Write(packet.Data()); err != nil {
+				log.Fatal(fmt.Errorf("main: writing packet failed: %w", err))
+			}
+
```
asticode, line 169: I think it's better to log success here
```go
log.Println("new packet")
```

asticode (changes requested):
Thanks for sticking with this PR! 👍

asticode:
@l0rem1psum we can also remove the example from the PR and merge it like this, and I can add the example later 👍

asticode:
FYI I've picked up the PR and merged it 👍

PR #71, opened 2024-09-05
Add audio FIFO

hsnks100:
I added an audio FIFO to the astiav library to support audio transcoding, such as AAC to Opus and Opus to AAC.

### DIFF
--- a/audio_fifo.go
+++ b/audio_fifo.go
@@ -1,0 +1,46 @@
+package astiav
+
+//#cgo pkg-config: libavutil
+//#include <libavutil/audio_fifo.h>
+//#include <stdlib.h>
+import "C"
+import "unsafe"
+
+type AudioFifo struct {
+	c *C.struct_AVAudioFifo
+}
+
+func newAudioFifoFromC(c *C.struct_AVAudioFifo) *AudioFifo {
+	if c == nil {
+		return nil
+	}
+	return &AudioFifo{c: c}
+}
+
+func AllocAudioFifo(sampleFmt SampleFormat, channels int, nbSamples int) *AudioFifo {
+	return newAudioFifoFromC(C.av_audio_fifo_alloc(C.enum_AVSampleFormat(sampleFmt), C.int(channels), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoRealloc(nbSamples int) int {
+	return int(C.av_audio_fifo_realloc((*C.struct_AVAudioFifo)(a.c), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoSize() int {
+	return int(C.av_audio_fifo_size((*C.struct_AVAudioFifo)(a.c)))
+}
+
+func (a *AudioFifo) AudioFifoSpace() int {
+	return int(C.av_audio_fifo_space((*C.struct_AVAudioFifo)(a.c)))
+}
+
+func (a *AudioFifo) AudioFifoWrite(data **uint8, nbSamples int) int {
+	return int(C.av_audio_fifo_write((*C.struct_AVAudioFifo)(a.c), (*unsafe.Pointer)(unsafe.Pointer(data)), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoRead(data **uint8, nbSamples int) int {
+	return int(C.av_audio_fifo_read((*C.struct_AVAudioFifo)(a.c), (*unsafe.Pointer)(unsafe.Pointer(data)), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoFree() {
+	C.av_audio_fifo_free((*C.struct_AVAudioFifo)(a.c))
+}
--- a/audio_fifo.go
+++ b/audio_fifo.go
@@ -34,17 +34,19 @@ func (a *AudioFifo) Space() int {
 }
 
 func (a *AudioFifo) Write(f *Frame) (int, error) {
-	frameRawData := (**uint8)(unsafe.Pointer(&f.c.data[0]))
-	ret := int(C.av_audio_fifo_write(a.c, (*unsafe.Pointer)(unsafe.Pointer(frameRawData)), C.int(f.NbSamples())))
+	ret := int(C.av_audio_fifo_write(a.c, (*unsafe.Pointer)(unsafe.Pointer(&f.c.data[0])), C.int(f.NbSamples())))
 	if ret < 0 {
 		return ret, newError(C.int(ret))
 	}
 	return ret, nil
 }
 
-func (a *AudioFifo) Read(f *Frame) int {
-	frameRawData := (**uint8)(unsafe.Pointer(&f.c.data[0]))
-	return int(C.av_audio_fifo_read(a.c, (*unsafe.Pointer)(unsafe.Pointer(frameRawData)), C.int(f.NbSamples())))
+func (a *AudioFifo) Read(f *Frame) (int, error) {
+	ret := int(C.av_audio_fifo_read(a.c, (*unsafe.Pointer)(unsafe.Pointer(&f.c.data[0])), C.int(f.NbSamples())))
+	if ret < 0 {
+		return ret, newError(C.int(ret))
+	}
+	return ret, nil
 }
 
 func (a *AudioFifo) Free() {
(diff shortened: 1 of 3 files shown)


audio_fifo.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return &AudioFifo{c: c}
+}
+
+func AllocAudioFifo(sampleFmt SampleFormat, channels int, nbSamples int) *AudioFifo {
+	return newAudioFifoFromC(C.av_audio_fifo_alloc(C.enum_AVSampleFormat(sampleFmt), C.int(channels), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoRealloc(nbSamples int) int {
+	return int(C.av_audio_fifo_realloc((*C.struct_AVAudioFifo)(a.c), C.int(nbSamples)))
```
asticode, line 25: Can you replace all the `(*C.struct_AVAudioFifo)(a.c)` in this file with only `a.c`?

audio_fifo.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (a *AudioFifo) AudioFifoSize() int {
+	return int(C.av_audio_fifo_size((*C.struct_AVAudioFifo)(a.c)))
+}
+
+func (a *AudioFifo) AudioFifoSpace() int {
+	return int(C.av_audio_fifo_space((*C.struct_AVAudioFifo)(a.c)))
+}
+
+func (a *AudioFifo) AudioFifoWrite(data **uint8, nbSamples int) int {
```
asticode, line 36: Right now the only way to get data plane pointers is using a frame, therefore can you replace `data **uint8, nbSamples int` arguments with `f *Frame` instead? Would that fit your use case as well?

audio_fifo.go
```diff
@@ -0,0 +1,46 @@
...
+	return int(C.av_audio_fifo_write((*C.struct_AVAudioFifo)(a.c), (*unsafe.Pointer)(unsafe.Pointer(data)), C.int(nbSamples)))
+}
+
+func (a *AudioFifo) AudioFifoRead(data **uint8, nbSamples int) int {
```
asticode, line 40: Can you also replace `data **uint8, nbSamples int` arguments with `f *Frame` instead?

frame.go
```diff
@@ -217,3 +217,7 @@ func (f *Frame) MoveRef(src *Frame) {
 func (f *Frame) UnsafePointer() unsafe.Pointer {
 	return unsafe.Pointer(f.c)
 }
+
+func (f *Frame) DataPtr() **uint8 {
```
asticode, line 221: There's no need for this function anymore 👍

audio_fifo.go
asticode, line 36: Can you return `(int, error)` instead of `int`?

For that you'll need to do something like:
```go
ret := C.av_audio_fifo_write(...)
if ret < 0 {
  return 0, newError(ret)
}
return ret, nil
```

audio_fifo.go
asticode, line 40: Can you return `(int, error)` instead of `int`?

For that you'll need to do something like:
```go
ret := C.av_audio_fifo_read(...)
if ret < 0 {
  return 0, newError(ret)
}
return ret, nil
```

audio_fifo_test.go
```diff
@@ -0,0 +1,36 @@
+package astiav
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestAudioFIFO(t *testing.T) {
+	audioFifo := AllocAudioFifo(
```
asticode, line 10: Can you rename `audioFifo` to `af`?

asticode, line 10: Can you add `defer af.Free()` below

audio_fifo_test.go
```diff
@@ -0,0 +1,36 @@
...
+	readFrame.SetChannelLayout(ChannelLayoutStereo)
+	readFrame.SetSampleFormat(SampleFormatFltp)
+	readFrame.SetSampleRate(48000)
+	readFrame.AllocBuffer(0)
+
+	written := audioFifo.AudioFifoWrite(writeFrame.DataPtr(), writeFrame.NbSamples())
+	require.Equal(t, writeSamples, written)
+	read := audioFifo.AudioFifoRead(readFrame.DataPtr(), readFrame.NbSamples())
+	require.Equal(t, readSamples, read)
+	require.Equal(t, audioFifo.AudioFifoSize(), writeSamples-readSamples)
```
asticode, line 35: Can you test `Realloc()` and `Space()` as well?

asticode (changes requested):
Thanks for the great PR!

~~Can you tell me more about your use case and why you needed `AudioFifo`?~~ (never mind, you described it in your first comment 👍 )

hsnks100:
I applied your all requests.

frame.go
asticode, line 221: Can you remove this function?

audio_fifo.go
```diff
@@ -0,0 +1,52 @@
...
+func (a *AudioFifo) Size() int {
+	return int(C.av_audio_fifo_size(a.c))
+}
+
+func (a *AudioFifo) Space() int {
+	return int(C.av_audio_fifo_space(a.c))
+}
+
+func (a *AudioFifo) Write(f *Frame) (int, error) {
+	frameRawData := (**uint8)(unsafe.Pointer(&f.c.data[0]))
```
asticode, line 37: Can you avoid creating this variable and use `(*unsafe.Pointer)(unsafe.Pointer(&f.c.data[0]))` instead of `(*unsafe.Pointer)(unsafe.Pointer(frameRawData))`?

audio_fifo.go
```diff
@@ -0,0 +1,52 @@
...
+	ret := int(C.av_audio_fifo_write(a.c, (*unsafe.Pointer)(unsafe.Pointer(frameRawData)), C.int(f.NbSamples())))
+	if ret < 0 {
+		return ret, newError(C.int(ret))
+	}
+	return ret, nil
+}
+
+func (a *AudioFifo) Read(f *Frame) int {
+	frameRawData := (**uint8)(unsafe.Pointer(&f.c.data[0]))
```
asticode, line 46: Can you avoid creating this variable and use `(*unsafe.Pointer)(unsafe.Pointer(&f.c.data[0]))` instead of `(*unsafe.Pointer)(unsafe.Pointer(frameRawData))`?

audio_fifo.go
```diff
@@ -0,0 +1,52 @@
...
+	return int(C.av_audio_fifo_read(a.c, (*unsafe.Pointer)(unsafe.Pointer(frameRawData)), C.int(f.NbSamples())))
```
asticode, line 47: Can you return `(int, error)` for this function the same way you did for the `Write()` method?

audio_fifo_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	readFrame := AllocFrame()
+	readFrame.SetNbSamples(readSamples)
+	readFrame.SetChannelLayout(ChannelLayoutStereo)
+	readFrame.SetSampleFormat(SampleFormatFltp)
+	readFrame.SetSampleRate(48000)
+	readFrame.AllocBuffer(0)
+
+	written, err := af.Write(writeFrame)
+	require.Equal(t, err, nil)
```
asticode, line 32: Can you replace this with `require.NoError(t, err)`?

audio_fifo_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.Equal(t, writeSamples, written)
+	read := af.Read(readFrame)
+	require.Equal(t, readSamples, read)
+	require.Equal(t, af.Size(), writeSamples-readSamples)
+	reallocSamples := 3000
+	err = af.Realloc(reallocSamples)
+	require.Equal(t, err, nil)
```
asticode, line 39: Can you replace this with `require.NoError(t, err)`?

asticode (changes requested):
Almost there 👍

asticode:
Great! I'll merge the PR and do some minor tweaks to avoid bothering you too much 👍 

Let me know whether you need a tag

hsnks100:
I need new version tag. Advance thank you.

asticode:
I've created the `v0.19.0` tag 👍

PR #73, opened 2024-09-21
Support for io.EOF to C.AVERROR_EOF mapping

daniel-sullivan:
At present, there's no mapping between the io.EOF returned by os.File or most io.Reader implementations and the astiav / ffmpeg ErrEof (C.AVERROR_EOF).

This PR adds a mapping in the event io.EOF is returned by the custom reader implementation.

### DIFF
--- a/io_context.go
+++ b/io_context.go
@@ -6,6 +6,7 @@ import "C"
 import (
 	"errors"
 	"fmt"
+	"io"
 	"sync"
 	"unsafe"
 )
@@ -241,6 +242,8 @@ func goAstiavIOContextReadFunc(opaque unsafe.Pointer, buf *C.uint8_t, bufSize C.
 		var e Error
 		if errors.As(err, &e) {
 			return C.int(e)
+		} else if errors.Is(err, io.EOF) {
+			return C.int(C.AVERROR_EOF)
 		}
 		return C.AVERROR_UNKNOWN
 	}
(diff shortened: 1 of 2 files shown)


io_context_test.go
```diff
@@ -9,34 +10,48 @@ import (
...
+		require.Equal(t, rb, b[:n])
+		_, err = c.Seek(2, 0)
+		require.NoError(t, err)
+		require.True(t, seeked)
+		c.Write(wb)
+		c.Flush()
+		require.Equal(t, wb, written)
+	})
+
+	t.Run("support go eof -> ffmpeg eof when reading", func(t *testing.T) {
```
asticode, line 44: Could you rename the test run to `io.EOF is mapped to AVERROR_EOF when reading`

asticode (changes requested):
Thanks for the PR!

Adding this mapping was my first instinct but I don't remember why I decided to remove it. Now this feels like a good idea 👍

io_context.go
```diff
@@ -241,6 +242,8 @@ func goAstiavIOContextReadFunc(opaque unsafe.Pointer, buf *C.uint8_t, bufSize C.
 		var e Error
 		if errors.As(err, &e) {
 			return C.int(e)
+		} else if errors.Is(err, io.EOF) {
+			return C.int(C.AVERROR_EOF)
```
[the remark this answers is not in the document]
daniel-sullivan: Fixed!

io_context_test.go
daniel-sullivan: Fixed!

asticode:
Thanks again for the PR ❤️ 

daniel-sullivan:
No worries! If I get a chance, I'll try and contribute an example for the audio FIFO implementation. Thank you for the great library!

No tag needed for now - we have a pretty easygoing policy of allowing go.mod to use commits rather than tags.

PR #84, opened 2024-10-14
Check if the frame is writable before copyPlanes

kurochan:
For safety of frame operations, check if the frame is writable before copyPlanes.

If this seems to be a performance overhead, I would like input as this PR may not be necessary.

### DIFF
--- a/frame_data.go
+++ b/frame_data.go
@@ -284,6 +284,11 @@ func (f *frameDataFrame) copyPlanes(ps []frameDataPlane) (err error) {
 	switch {
 	// Video
 	case f.height() > 0 && f.width() > 0:
+		// Check writability
+		if !f.f.IsWritable() {
+			return errors.New("astiav: frame is not writable")
+		}
+
 		// Loop through planes
 		var cdata [8]*C.uint8_t
 		var clinesizes [8]C.int
--- a/frame_data.go
+++ b/frame_data.go
@@ -280,15 +280,15 @@ func (f *frameDataFrame) bytes(align int) ([]byte, error) {
 	return nil, errors.New("astiav: frame type not implemented")
 }
 
-func (f *frameDataFrame) copyPlanes(ps []frameDataPlane) (err error) {
+func (f *frameDataFrame) copyPlanes(ps []frameDataPlane) error {
+	// Check writability
+	if !f.f.IsWritable() {
+		return errors.New("astiav: frame is not writable")
+	}
+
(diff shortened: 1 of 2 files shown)


frame_data.go
```diff
@@ -284,6 +284,11 @@ func (f *frameDataFrame) copyPlanes(ps []frameDataPlane) (err error) {
 	switch {
 	// Video
 	case f.height() > 0 && f.width() > 0:
+		// Check writability
```
asticode, line 287: Could you check the writability above the `switch`?

frame_data.go
```diff
@@ -284,6 +284,11 @@ func (f *frameDataFrame) copyPlanes(ps []frameDataPlane) (err error) {
 	switch {
```
asticode, line 284: Could you replace `(err error)` with `error`?

asticode (changes requested):
At first I forced the frame to be writable by adding `.MakeWritable()` when copying planes but then I realized the developer should be free to decide when to make the frame writable. However checking whether the frame is writable seems like a good idea even though I don't really know whether this is a big performance drop 👍 

Could you also add tests for this use case? Something like the following added [here](https://github.com/asticode/go-astiav/blob/master/frame_data_test.go#L521)(after a newline)
```go
f3 := astiav.AllocFrame()
defer f3.Free()
require.NoError(t, f3.Ref(f2))
require.Error(t, fd2.FromImage(i1))
require.Error(t, fd2.SetBytes(b1, align))
```

frame_data_test.go
```diff
@@ -518,5 +518,14 @@ func TestFrameData(t *testing.T) {
 		require.NoError(t, err)
 		b9 := []byte(fmt.Sprintf("%+v", b8))
 		require.Equal(t, b3, b9)
+
+		f3 := AllocFrame()
+		defer f3.Free()
+		require.NoError(t, f3.Ref(f2))
+		require.Error(t, fd2.FromImage(i1))
+		require.Error(t, fd2.SetBytes(b1, align))
+		f3.MakeWritable()
```
asticode, line 527: `f2` should be the frame that we make writable here

asticode (changes requested):

asticode:
Thanks for the PR! ❤️ 

kurochan:
Please tag this as well 🚀

asticode:
I've created a `v0.23.0` tag 👍

PR #94, opened 2024-11-04
Create hardware device context with flags

l0rem1psum:
Despite the official doc claiming that the flags are not used (see: [hwcontext.c File Reference](https://www.ffmpeg.org/doxygen/3.2/hwcontext_8c.html#a21fbd088225e4e25c4d9a01b3f5e8c51)), it is in reality used by `AVCUDADeviceContext` (see: [hwcontext_cuda.h](https://github.com/FFmpeg/FFmpeg/blob/df00705e0010cc2c53d17d51944f847c2c852189/libavutil/hwcontext_cuda.h#L53)).

### DIFF
--- a/examples/hardware_decoding/main.go
+++ b/examples/hardware_decoding/main.go
@@ -133,7 +133,7 @@ func main() {
 
 		// Create hardware device context
 		var err error
-		if s.hardwareDeviceContext, err = astiav.CreateHardwareDeviceContext(hardwareDeviceType, *hardwareDeviceName, nil); err != nil {
+		if s.hardwareDeviceContext, err = astiav.CreateHardwareDeviceContext(hardwareDeviceType, *hardwareDeviceName, nil, 0); err != nil {
 			log.Fatal(fmt.Errorf("main: creating hardware device context failed: %w", err))
 		}
 
--- a/examples/hardware_encoding/main.go
+++ b/examples/hardware_encoding/main.go
@@ -49,7 +49,7 @@ func main() {
 	}
 
 	// Create hardware device context
-	hardwareDeviceContext, err := astiav.CreateHardwareDeviceContext(hardwareDeviceType, *hardwareDeviceName, nil)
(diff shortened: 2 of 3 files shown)


asticode:
I've added flags this way in the lib (example is for codec context flags):

- list of flag values: https://github.com/asticode/go-astiav/blob/master/codec_context_flag.go
- struct capable of manipulating flag values: https://github.com/asticode/go-astiav/blob/master/flags.go#L48 (this file is generated automatically with [this command](https://github.com/asticode/go-astiav/blob/master/internal/cmd/flags/main.go))
- using the flag: https://github.com/asticode/go-astiav/blob/master/codec_context.go#L127

We should do the same thing for the new flags you're mentioning.

It would mean:

- create a `HardwareDeviceContextFlag` type in a new `hardware_device_context_flag.go` file with both `AV_CUDA_USE_PRIMARY_CONTEXT` and `AV_CUDA_USE_CURRENT_CONTEXT` as possible values with go constants named `HardwareDeviceContextFlagCudaUsePrimaryContext` and `HardwareDeviceContextFlagCudaUseCurrentContext`
- modify the generate flags command to generate a `HardwareDeviceContextFlags` struct and run it to actually add it to `flags.go`
- use `HardwareDeviceContextFlags` instead of `int` in `CreateHardwareDeviceContext`

l0rem1psum:
Thanks for the reply. My main concern is such a flag (`AV_CUDA_USE_CURRENT_CONTEXT` in this case) wouldn't be available to users who didn't build ffmpeg with CUDA, rendering this binding strongly coupled to a particular hardware device.

As such, I still strongly recommend using a `int` flag and avoid adding custom types not available in the C code. (And avoid adding additional logic in FFI bindings in general.)

l0rem1psum:
I just added the suggested code but it wouldn't compile without linking CUDA. Hence my above argument still stands.

asticode:
You're right, unfortunately I don't see a workaround for this 🤔 

Thanks for this PR too! ❤️

PR #95, opened 2024-11-04
Expose pointer to underlying AVCodecContext

l0rem1psum:
In fact, I propose adding similar methods to all C wrapper structs for better interoperability with client codes. Also, a reciprocal method/function to convert client's `unsafe.Pointer`s to their corresponding `go-astiav` bindings.


asticode:
Good idea, could you also add a test similar to [this](https://github.com/asticode/go-astiav/blob/master/frame_test.go#L21)?

asticode:
Thanks for the PR ❤️ 

PR #111, opened 2024-11-25
codec_context.go and format_context.go

oldma3095:
codec_context.go
`
func (cc *CodecContext) HardwareFrameContext() *HardwareFrameContext{
	return newHardwareFrameContextFromC(cc.c.hw_frames_ctx)
}
`
format_context.go
`
func (fc *FormatContext) DumpFormat(index int32, url string, isOutput int32) {
	urlc := (*C.char)(nil)
	if len(url) > 0 {
		urlc = C.CString(url)
		defer C.free(unsafe.Pointer(urlc))
	}
	C.av_dump_format(fc.c, C.int(index), urlc, C.int(isOutput))
}
`

### DIFF
--- a/format_context.go
+++ b/format_context.go
@@ -366,13 +366,11 @@ func (fc *FormatContext) FindBestStream(mt MediaType, wantedStreamIndex, related
 }
 
 // https://ffmpeg.org/doxygen/7.0/group__lavf__misc.html#gae2645941f2dc779c307eb6314fd39f10
-// prints detailed information about the input or output format, such as
-// duration, bitrate, streams, container, programs, metadata, side data, codec and time base.
 func (fc *FormatContext) DumpFormat(index int32, url string, isOutput int32) {
-	urlc := (*C.char)(nil)
+	curl := (*C.char)(nil)
 	if len(url) > 0 {
-		urlc = C.CString(url)
-		defer C.free(unsafe.Pointer(urlc))
+		curl = C.CString(url)
+		defer C.free(unsafe.Pointer(curl))
 	}
-	C.av_dump_format(fc.c, C.int(index), urlc, C.int(isOutput))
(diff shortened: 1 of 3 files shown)


format_context.go
```diff
@@ -364,3 +364,15 @@ func (fc *FormatContext) FindBestStream(mt MediaType, wantedStreamIndex, related
 	}
 	return nil, nil, fmt.Errorf("astiav: no stream with index %d", ret)
 }
+
+// https://ffmpeg.org/doxygen/7.0/group__lavf__misc.html#gae2645941f2dc779c307eb6314fd39f10
+// prints detailed information about the input or output format, such as
+// duration, bitrate, streams, container, programs, metadata, side data, codec and time base.
+func (fc *FormatContext) DumpFormat(index int32, url string, isOutput int32) {
```
asticode, line 371: Could you rename the method to `Dump` and update arguments to `streamIndex int, url string, isOutput bool`?

format_context.go
```diff
@@ -364,3 +364,15 @@ func (fc *FormatContext) FindBestStream(mt MediaType, wantedStreamIndex, related
...
+	urlc := (*C.char)(nil)
```
asticode, line 372: To follow the same convention as in the rest of the lib, could you rename this variable to `curl` instead of `urlc`?

format_context.go
asticode, line 371: Can you add a test [here](https://github.com/asticode/go-astiav/blob/master/format_context_test.go#L37) (keep a newline before and after your new test) that would set the log callback, dump the format context, assert the content of the logs **and reset the log callback**?

asticode (changes requested):
Thanks for the PR!

asticode:
Thanks for the PR! ❤️ 

I'll do some minor tweaks in the test in the coming minutes, but let me know whether you need a tag 👍

PR #113, opened 2024-11-27
av_q2d

oldma3095:
mathematics.go
mathematics_test.go


asticode:
[Rational.Float64()](https://github.com/asticode/go-astiav/blob/master/rational.go#L39) already exists, isn't it doing the exact same thing?

PR #116, opened 2024-12-10
OpenIOContextWithDictionary

oldma3095:
need open output with dictionary.
example:

d := astiav.NewDictionary()
_ = d.Set("pkt_size", "1316", 0)
ctx, err := astiav.OpenIOContextWithDictionary("udp://239.0.0.1:12345", astiav.NewIOContextFlags(astiav.IOContextFlagWrite), d)


asticode:
I'd rather keep only one function opening an io context therefore could you modify `OpenIOContext` directly by updating the link in its comment, adding the dictionnary as an argument and the logic inside of it? Also, can you modify examples that use this function?

io_context.go
```diff
@@ -108,12 +108,16 @@ func AllocIOContext(bufferSize int, writable bool, readFunc IOContextReadFunc, s
 	return
 }
 
-// https://ffmpeg.org/doxygen/7.0/avio_8c.html#ab1b99c5b70aa59f15ab7cd4cbb40381e
-func OpenIOContext(filename string, flags IOContextFlags) (*IOContext, error) {
+// https://ffmpeg.org/doxygen/7.0/avio_8c.html#ae8589aae955d16ca228b6b9d66ced33d
+func OpenIOContext(filename string, flags IOContextFlags, d *Dictionary) (*IOContext, error) {
```
asticode, line 112: Can you also add `ii *IOInterrupter` as an argument and provide it to `avio_iopen2`?
```go
var cii *C.AVIOInterruptCB = nil
if ii != nil {
  cii = &ii.c
}
C.avio_open2(&c, cfi, C.int(flags), cii, dc)
```
Don't forget to update examples and other places where it's used 👍

io_context_test.go
```diff
@@ -56,7 +56,9 @@ func TestIOContext(t *testing.T) {
 
 func TestOpenIOContext(t *testing.T) {
 	path := filepath.Join(t.TempDir(), "iocontext.txt")
-	c, err := OpenIOContext(path, NewIOContextFlags(IOContextFlagWrite))
+	d := NewDictionary()
```
asticode, line 59: I don't really see the point of providing a real `Dictionary` here since we're not asserting anything linked to it below, therefore you can remove
```go
d := NewDictionary()
_ = d.Set("k", "v", 0)
```

asticode (changes requested):

asticode:
Thanks for the PR! ❤️

PR #125, opened 2024-12-23
gragh multi filters set hardware device context

oldma3095:
FilterGraph.NbFilters()
FilterGraph.Filters()
FilterContext.SetHardwareDeviceContext(hdc *HardwareDeviceContext)

### DIFF
--- a/filter_graph_test.go
+++ b/filter_graph_test.go
@@ -144,6 +144,7 @@ func TestFilterGraph(t *testing.T) {
 
 		buffersinkContext, err := fg.NewBuffersinkFilterContext(buffersink, "filter_out")
 		require.NoError(t, err)
+		require.Equal(t, buffersink, buffersinkContext.FilterContext().Filter())
 		cl = buffersinkContext.FilterContext().Class()
 		require.NotNil(t, cl)
 		require.Equal(t, "AVFilter", cl.Name())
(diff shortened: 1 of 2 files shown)


filter_graph_test.go
```diff
@@ -247,6 +247,43 @@ func TestFilterGraph(t *testing.T) {
 	require.Equal(t, 1, len(os))
 	require.Equal(t, MediaTypeAudio, os[0].MediaType())
 
+	fg3 := AllocFilterGraph()
```
asticode, line 250: I don't really see the point of adding this, it's redundant with [this](https://github.com/asticode/go-astiav/blob/master/filter_graph_test.go#L217).

Instead could you add [here](https://github.com/asticode/go-astiav/blob/master/filter_graph_test.go#L196) something like:
```go
require.Equal(t, len(buffersrcContexts)+1, fg.NbFilters())
fs := fg.Filters()
require.Equal(t, len(buffersrcContexts)+1, len(fs))
require.Equal(t, buffersinkContext.FilterContext(), fs[0])
for idx, c := range fs[1:] {
  require.Equal(t, buffersrcContexts[idx].FilterContext(), c)
}
```
I don't know whether the buffersink context comes first in `fg.Filters()` therefore you may have to reverse order when testing it.

asticode:
I've added `FilterFlag` in order to add the [logic](https://github.com/FFmpeg/FFmpeg/blob/n7.0/fftools/ffmpeg_filter.c#L577) you linked in the previous PR once this PR is merged, could you fix conflicts and resolve my comment?

oldma3095:
i have merged master `FilterFlag` and create a new PR

PR #127, opened 2025-01-06
Frame.Copy(dst *Frame)

oldma3095:


frame_test.go
```diff
@@ -55,11 +55,19 @@ func TestFrame(t *testing.T) {
...
+	f33 := f3.Clone()
+	err = f3.Copy(f33)
+	require.NoError(t, err)
+	defer f33.Free()
+
 	err = f2.AllocBuffer(0)
 	require.NoError(t, err)
 	err = f3.Ref(f2)
 	require.NoError(t, err)
-	require.Equal(t, 2, f3.Height())
```
asticode, line 62: Can you remove this change?

frame_test.go
```diff
@@ -55,11 +55,19 @@ func TestFrame(t *testing.T) {
...
+	require.Contains(t, [][8]int{
```
asticode, line 67: Can you remove this change?

frame_test.go
```diff
@@ -55,11 +55,19 @@ func TestFrame(t *testing.T) {
 	defer f3.Free()
 	require.Equal(t, 180, f3.Height())
 
+	f33 := f3.Clone()
```
asticode, line 58: Can you do that instead and move the test to the bottom of the test function?

```go
f7 := AllocFrame()
require.NotNil(t, f7)
defer f7.Free()
align = 1
f7.SetHeight(f1.Height())
f7.SetPixelFormat(f1.PixelFormat())
f7.SetWidth(f1.Width())
require.NoError(t, f7.AllocBuffer(align))
require.NoError(t, f7.AllocImage(align))
require.NoError(t, f1.Copy(f7))
require.Equal(t, f1.Data().Bytes(), f7.Data().Bytes())
```

asticode (changes requested):
Thanks for the PR!

A few changes needed in the tests 👍

frame_test.go
oldma3095: image
image

`f3.MoveRef(f1)` ,`f1` is a empty `Frame`.
test can not pass.

asticode: Fair enough, can you try using `f6` instead of `f1` then?

oldma3095:
if u merge it, can u release a new tag?

asticode:
Thanks again for the PR! ❤️ 

I've created a `v0.30.0` tag 👍

PR #129, opened 2025-01-13
stream: Add functions to get/set fields of Stream

xaionaro:
Just adding few missing Getters/Setters to `Stream`.


stream.go
```diff
@@ -120,3 +130,15 @@ func (s *Stream) TimeBase() Rational {
 func (s *Stream) SetTimeBase(r Rational) {
 	s.c.time_base = r.c
 }
+
+type PTSWrapBits int
```
asticode, line 134: Can you remove this type? No need for it, you can use the `int` type directly

stream.go
asticode, line 1: Can you add tests as well [here](https://github.com/asticode/go-astiav/blob/master/stream_test.go#L39) for all getters/setters you've added, ordered alphabetically in the form
```go
s1.SetEventFlags(1)
require.Equal(t, 1, s1.EventFlags())
```

asticode (changes requested):
Thanks for the PR!

A few changes needed 👍

stream.go
xaionaro: Done.

stream.go
xaionaro: Done.

asticode:
Thanks again for the PR! ❤️ 

PR #130, opened 2025-01-17
Improved avdevice support

peakedshout:
Hello! I've been using this library for a while and I'm happy with the go binding!
I used to use gmf, and the methods it provides are modified so I don't have a good idea what I'm calling? And no one seems to be hot maintenance.
In my confusion, I found this library.
I found a method called RegisterAllDevices(). I happen to be doing some device monitoring at the moment, so I need some extra methods.
I checked InputAudioDeviceNext, OutputAudioDeviceNext, InputVideoDeviceNext, OutputVideoDeviceNext.
As for the DeviceInfo section, I also checked it from the documentation and it worked "just fine". It says "Function not implemented", which seems to be an implementation issue with ffmpeg, but interfacing shouldn't be too problematic.
About the AVAppToDevMessageType part, I'm not familiar with this part, maybe I'll explore it later.

Please let me know if there are any changes that need to be made.


asticode:
Thanks for contributing!

Before digging into reviewing your PR, I'm not looking to add functions that you're not going to use actively. That would complicate maintaining this library for nothing and I don't want that.

Therefore could you list the methods you've added that you are **really** using?

asticode:
I'm welcoming your PR but I need to understand better what you really need in order to provide the best API for this lib.

> // 1. Gets the available input audio devices
	list := astiav.InputAudioDeviceList()
	for _, format := range list {
		fmt.Println(format.Name())
	}

In your example you're simply using `fmt.Println(format.Name())` but I guess in your use case you're going to use it another way. In which precise way are you going to use it?

> // 2. Gets the default input audio device
	cx := astiav.AllocFormatContext()
	_ = cx.OpenInput("default", astiav.InputAudioDeviceNext(nil), nil)

Are you only looking for the default input audio device in your use case? Or do you also need to get another of the input audio device in the list?

> // 3. Gets the available input source device information （This is supported by the ffmpeg implementation）
	sources, _ := astiav.ListInputSources(astiav.InputAudioDeviceNext(nil), "", nil)
	for _, info := range sources.Devices() {
		fmt.Println(info.DeviceName())
	}

Why do you need this? In your example, aren't `astiav.InputAudioDeviceList()` or `astiav.InputAudioDeviceNext(nil)` enough?

> // 4. Other similar ones

Can you elaborate and provide at least a list of method names so that I can understand better?

peakedshout:
I'm not sure what you want me to explain specifically. These methods are used in my program to implement these methods to obtain device information. (If the operating environment supports it)

1. Use this method to scan the driver device.
RegisterAllDevices
2. These methods give the corresponding Format, and I will use these to open the device (when the parameter is nil, the first device will be returned).
InputAudioDeviceNext
OutputAudioDeviceNext
InputVideoDeviceNext
OutputVideoDeviceNext
3. These are all Formats. This is a simple encapsulation method to output which Formats the operating environment can use.
InputAudioDeviceList
OutputAudioDeviceList
InputVideoDeviceList
OutputVideoDeviceList
4. These are essentially a method, but ffmpeg provides some convenient calling methods to return specific driver device information. If the operating environment supports it, I will also display this device information for the user to view.
(fc *FormatContext) ListDevices ListInputSources ListOutputSinks

asticode:
OK thanks for providing all this information 👍 

It looks like if we were to add the following that would fulfill your needs:

```go
func InputAudioDevices() []*InputFormat
func InputVideoDevices() []*InputFormat
func OutputAudioDevices() []*OutputFormat
func OutputVideoDevices() []*OutputFormat

type DeviceInfo struct
func (i DeviceInfo) Name() string
func (i DeviceInfo) Description() string
func (i DeviceInfo) MediaTypes() []MediaType

type DeviceInfoList struct
func (l *DeviceInfoList) List() []*DeviceInfo
func (l *DeviceInfoList) Default() *DeviceInfo
func (l *DeviceInfoList) Free()

func (fc *FormatContext) Devices() (*DeviceInfoList, error)
func InputSources(device *InputFormat, deviceName string, deviceOptions *Dictionary) (*DeviceInfoList, error)
func OutputSinks(device *OutputFormat, deviceName string, deviceOptions *Dictionary) (*DeviceInfoList, error)
```

Can you confirm?

peakedshout:
I'm a little bit skeptical, isn't the source method preserved? And don't they follow the original name?
In my opinion, keeping some degree of source methods is easier to maintain.
Also, the name is closer to the original name, which makes it easier to understand what you are calling ffmpeg.
(Some of my experiences with some binding libraries: I knew what the source method was, but I had no idea what it was replaced with because the name had changed.)

asticode:
I'd rather not mirror the FFmpeg naming in Go since it can be a bit esoteric sometimes and not very fitting in the Go world. The naming I prefer may not be best in all cases, but for those discovering FFmpeg I've been told it's way easier to get into it. For those already knowing FFmpeg they usually have the skillset to look for the C methods in the Go code 👍

peakedshout:
No, I have used it for a while, and I think the PR naming I submitted can well express what to execute. And each method is in use, which can help me control some things during development.
And I have some opinions: for some hardware and deep-level methods, keeping direct expressions will help the development process very well. As you can see, I will consult the documentation during development, and I will be troubled when I see strange naming (and the premise is that there is documentation).
I don’t deny that some simplified encapsulation will facilitate development work, but it should not be completely simplified.

asticode:
I understand your point however I'm not looking to build the API of this lib this way. Feel free to close this PR if you don't feel like making the requested changes 👍

peakedshout:
Unfortunately, it looks like I'll just have to fork the repository and continue my work.

PR #146, opened 2025-04-17
Add AV1 codec id

3DRX:
@asticode Hi first of all, thanks for this great library! I'm currently using this to decode av1 video stream, and I found that AV1 codec id is not included in `codec_id.go`.

Perhaps in the future, a script or something can be made to generate `codec_id.go` from `libavcodec/codec_id.h`? I'd love to do that if you think it would be helpful.

### DIFF
--- a/codec_id.go
+++ b/codec_id.go
@@ -413,6 +413,7 @@ const (
 	CodecIDZerocodec                = CodecID(C.AV_CODEC_ID_ZEROCODEC)
 	CodecIDZlib                     = CodecID(C.AV_CODEC_ID_ZLIB)
 	CodecIDZmbv                     = CodecID(C.AV_CODEC_ID_ZMBV)
+	CodecIDAv1                      = CodecID(C.AV_CODEC_ID_AV1)
 )
 
 func (c CodecID) MediaType() MediaType {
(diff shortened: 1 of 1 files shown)


codec_id.go
```diff
@@ -413,6 +413,7 @@ const (
 	CodecIDZerocodec                = CodecID(C.AV_CODEC_ID_ZEROCODEC)
 	CodecIDZlib                     = CodecID(C.AV_CODEC_ID_ZLIB)
 	CodecIDZmbv                     = CodecID(C.AV_CODEC_ID_ZMBV)
+	CodecIDAv1                      = CodecID(C.AV_CODEC_ID_AV1)
```
asticode, line 416: Could you order the new variable alphabetically?

asticode (changes requested):
Thanks for the PR ❤️ 

I don't believe a script is necessary for now but that's a good idea so I'll keep it in mind 👍

sylwek78 (approved):

sylwek78:
[]()

codec_id.go
3DRX: Sure!

asticode:
Thanks a lot for the PR ❤️ 

asticode:
FYI I've created the `v0.36.0` tag 👍

PR #154, opened 2025-05-23
Add AVPixFmtDescriptor to retrieve flags for AVPixelFormat

Maizer:
feat(PixelFormat): Expose `AVPixFmtDescriptor` via `Descriptor()` method

This commit introduces support for `AVPixFmtDescriptor`, which provides
detailed information about an `AVPixelFormat`, including its flags.

A new `Descriptor()` method has been added to the `PixelFormat` type
(or class) to retrieve its corresponding `AVPixFmtDescriptor`. This allows
for easier access to extended properties of pixel formats.


asticode (changes requested):
Thanks (again) for the great PR! ❤️ 

Great to see you could find how to generate flags 👍 

A few minor changes needed.

More importantly could you add tests:
- in `pixel_format_test.go` => test the `Descriptor()` method by only checking the returned descriptor is non nil and name is correct
- add a `pixel_format_descriptor_test.go` file which would test all methods of the new `PixelFormatDescriptor` struct

Maizer:
Alright, thank you for your guidance/correction. I've implemented the tests. As this is my first time [contributing tests like this], I'm not entirely sure if they're right, so I'd appreciate your review on them.

pixel_format_descriptor_test.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestPixelFormatDescriptor(t *testing.T) {
+	p := PixelFormatCuda
+
```
asticode, line 11: You can simplify with
```go
p := PixelFormatCuda
d := p.Descriptor()
require.NotNil(t, d)
require.Equal(t, d.Name() p.String())
require.True(t, d.Flags().Has(PixelFormatDescriptorFlagHwAccel))
```

(no need for empty lines in between)

pixel_format_test.go
```diff
@@ -10,4 +10,8 @@ func TestPixelFormat(t *testing.T) {
 	p := FindPixelFormatByName("yuv420p")
 	require.Equal(t, PixelFormatYuv420P, p)
 	require.Equal(t, "yuv420p", p.String())
+
```
asticode, line 13: You can remove the empty line

asticode (changes requested):
Your tests are good 👍  It needs minor changes but more for consistency with other tests in the lib

asticode:
Thanks again for the PR! ❤️ 

PR #161, opened 2025-07-14
Add json compatible encoding for core types

nveeser:
Update Rational, PixelFormat and ChannelLayout to implement json interfaces which allow serializing these types to JSON without intermediate parser steps

Also add names for levels so the level name can be printed using fmt.Printf("%s")

### DIFF
--- a/channel_layout.go
+++ b/channel_layout.go
@@ -3,6 +3,7 @@ package astiav
 //#include "channel_layout.h"
 import "C"
 import (
+	"fmt"
 	"unsafe"
 )
 
@@ -122,4 +123,26 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
 	err := l.copy(&cl)
 	dst := newChannelLayoutFromC(&cl)
 	return dst, err
+}
+
+func ParseChannelLayout(s string) (ChannelLayout, error) {
+	var c C.AVChannelLayout
+	cn := C.CString(s)
+	defer C.free(unsafe.Pointer(cn))
+	id := C.av_channel_layout_from_string(&c, cn)
+	return newChannelLayoutFromC(&c), newError(id)
+}
+
+func (l ChannelLayout) MarshalText() ([]byte, error) { return ([]byte)(l.String()), nil }
+func (l *ChannelLayout) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		return nil
+	}
+	pf, err := ParseChannelLayout(s)
+	if err != nil {
+		return fmt.Errorf("invalid Channel Layout: %s: %w", s, err)
+	}
+	l.c = pf.c
+	return nil
 }
--- a/log.go
+++ b/log.go
@@ -22,6 +22,27 @@ const (
 	LogLevelDebug   = LogLevel(C.AV_LOG_DEBUG)
 )
 
+func (l LogLevel) String() string {
+	switch l {
+	case LogLevelPanic:
+		return "PANIC"
+	case LogLevelFatal:
+		return "FATAL"
+	case LogLevelError:
+		return "ERROR"
+	case LogLevelWarning:
+		return "WARNING"
+	case LogLevelInfo:
+		return "INFO"
+	case LogLevelVerbose:
+		return "VERBOSE"
+	case LogLevelDebug:
+		return "DEBUG"
+	default:
+		return "UNKNOWN"
+	}
+}
+
 // https://ffmpeg.org/doxygen/7.0/group__lavu__log.html#ga1fd32c74db581e3e2e7f35d277bb1e24
 func SetLogLevel(l LogLevel) {
 	C.av_log_set_level(C.int(l))
--- a/pixel_format.go
+++ b/pixel_format.go
@@ -5,10 +5,17 @@ package astiav
 //#include <libavutil/pixdesc.h>
 //#include <libavutil/pixfmt.h>
 import "C"
-import "unsafe"
+import (
+	"encoding"
+	"fmt"
+	"unsafe"
+)
 
 // https://ffmpeg.org/doxygen/7.0/pixfmt_8h.html#a9a8e335cf3be472042bc9f0cf80cd4c5
 type PixelFormat C.enum_AVPixelFormat
+
+var _ encoding.TextMarshaler = PixelFormatNone
+var _ encoding.TextUnmarshaler = (*PixelFormat)(nil)
 
 const (
 	PixelFormat0Bgr          = PixelFormat(C.AV_PIX_FMT_0BGR)
@@ -220,4 +227,18 @@ func FindPixelFormatByName(name string) PixelFormat {
 	cn := C.CString(name)
 	defer C.free(unsafe.Pointer(cn))
 	return PixelFormat(C.av_get_pix_fmt(cn))
+}
+
+func (f PixelFormat) MarshalText() ([]byte, error) { return ([]byte)(f.String()), nil }
+func (f *PixelFormat) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		*f = PixelFormatNone
+	}
+	pf := FindPixelFormatByName(s)
+	if pf == PixelFormatNone {
+		return fmt.Errorf("invalid PixelForamt: %s", s)
+	}
+	*f = pf
+	return nil
 }
--- a/pixel_format_test.go
+++ b/pixel_format_test.go
@@ -7,10 +7,23 @@ import (
 )
 
 func TestPixelFormat(t *testing.T) {
-	p := FindPixelFormatByName("yuv420p")
-	require.Equal(t, PixelFormatYuv420P, p)
-	require.Equal(t, "yuv420p", p.String())
 	d := p.Descriptor()
 	require.NotNil(t, d)
 	require.Equal(t, d.Name(), p.String())
+	p := FindPixelFormatByName("yuv420p")
+	require.Equal(t, PixelFormatYuv420P, p)
+	require.Equal(t, "yuv420p", p.String())
+	t.Run("FindByName", func(t *testing.T) {
+		p := FindPixelFormatByName("yuv420p")
+		require.Equal(t, PixelFormatYuv420P, p)
+		require.Equal(t, "yuv420p", p.String())
+	})
+	t.Run("Encode", func(t *testing.T) {
+		x1 := PixelFormatAbgr
+		s, err := x1.MarshalText()
+		require.NoError(t, err)
+		var x2 PixelFormat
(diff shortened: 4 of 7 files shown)


nveeser:
This basically allows storing these values in a struct and serializing that struct to JSON or any config format (yaml, toml, etc) that supports encoding.TextMarshaler interface.

rational_test.go
```diff
@@ -1,23 +1,57 @@
...
+		require.Equal(t, 0.0, r.Float64())
+		r = NewRational(2, 1)
+		require.Equal(t, 2.0, r.Float64())
+		r = NewRational(1, 2)
+		require.Equal(t, 0.5, r.Float64())
+	})
+	t.Run("String", func(t *testing.T) {
+		r := NewRational(2, 1)
+		require.Equal(t, "2/1", r.String())
+	})
```
asticode, line 35: Can you also test `r.String()` returns `"0"` if you `SetDen(0)?`

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
 func (r Rational) Invert() Rational {
 	return NewRational(r.Den(), r.Num())
 }
+
+type rational struct {
+	Num int
+	Den int
+}
+
+func (r *Rational) UnmarshalJSON(d []byte) error {
```
asticode, line 73: Why do you need to add `UnmarshalJSON` and `MarshalJSON` explicitely here?

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
...
+	}
+	r.SetNum(rs.Num)
+	r.SetDen(rs.Den)
+	return nil
+}
+
+func (r Rational) MarshalJSON() ([]byte, error) {
+	return json.Marshal(rational{Num: r.Num(), Den: r.Den()})
+}
+
```
asticode, line 86: I'd rather add a method `NewRationalFromString(s string) (Rational, error)` that you can use in `UnmarshalText`

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
...
+func (r Rational) MarshalText() ([]byte, error) { return ([]byte)(r.String()), nil }
+func (r *Rational) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		r.SetNum(0)
```
asticode, line 91: Can you do `*r = NewRational(0, 0)` instead?

Also you need to return `nil` here.

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
...
+		r.SetDen(0)
+	}
+	parts := strings.Split(s, "/")
```
asticode, line 94: Here you need to call the new `NewRationalFromString` method (just like for the `ChannelLayout`)

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
...
+	if len(parts) != 2 {
```
asticode, line 95: Having one part is actually valid: `12` for instance is `12/1`. Therefore I'd rather have:
```go
if len(parts) < 1 {
 return errors.New("astiav: invalid number of parts")
}
num, err := strconv.ParseInt(parts[0], 10, 0)
if err != nil {
  return fmt.Errorf("astiav: parsing numerator failed: %w", err)
}
den := 1
if len(parts) > 1 {
  if den, err := strconv.ParseInt(parts[1], 10, 0); err != nil {
    return fmt.Errorf("astiav: parsing denumerator failed: %w", err)
  }
}
```

rational.go
```diff
@@ -53,3 +64,46 @@ func (r Rational) String() string {
...
+	}
+	num, err := strconv.ParseInt(parts[0], 10, 0)
+	if err != nil {
+		return fmt.Errorf("value %s is not Rational ('d/d'): %w", s, err)
+	}
+	den, err := strconv.ParseInt(parts[1], 10, 0)
+	if err != nil {
+		return fmt.Errorf("value %s is not Rational ('d/d'): %w", s, err)
+	}
+	r.SetNum(int(num))
```
asticode, line 106: I'd rather have `*r = NewRational(int(num), int(den))`

rational_test.go
```diff
@@ -1,23 +1,57 @@
...
+	t.Run("TextMarshal", func(t *testing.T) {
+		r := NewRational(2, 1)
+		s, err := r.MarshalText()
```
asticode, line 38: Can you rename `s` to `b`?

rational_test.go
```diff
@@ -1,23 +1,57 @@
...
+		require.NoError(t, err)
```
asticode, line 39: You need to add `require.Equal(t, "2/1", string(b))`

rational_test.go
```diff
@@ -1,23 +1,57 @@
...
+		var r2 Rational
+		require.NoError(t, r2.UnmarshalText(s))
+		require.Equal(t, r.Num(), r2.Num())
+		require.Equal(t, r.Den(), r2.Den())
+	})
+	t.Run("json.Marshal", func(t *testing.T) {
```
asticode, line 45: You don't need this if we determine json methods are not needed

asticode, line 45: You need to add tests for the new `NewRationalFromString` method

nveeser: Did I miss a comment?

Are you looking for `NewRationalFromString(string)  (Rational, error)`

channel_layout.go
```diff
@@ -123,3 +124,25 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
...
+
+func ParseChannelLayout(s string) (ChannelLayout, error) {
+	var c C.AVChannelLayout
+	cn := C.CString(s)
+	defer C.free(unsafe.Pointer(cn))
+	id := C.av_channel_layout_from_string(&c, cn)
+	return newChannelLayoutFromC(&c), newError(id)
+}
+
+func (l ChannelLayout) MarshalText() ([]byte, error) { return ([]byte)(l.String()), nil }
```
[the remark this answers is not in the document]
nveeser: added type assertions. I think I added them in the expected / consistent place?

pixel_format.go
```diff
@@ -5,11 +5,18 @@ package astiav
...
+import (
+	"encoding"
+	"fmt"
+	"unsafe"
+)
 
 // https://ffmpeg.org/doxygen/7.0/pixfmt_8h.html#a9a8e335cf3be472042bc9f0cf80cd4c5
 type PixelFormat C.enum_AVPixelFormat
 
+var _ encoding.TextMarshaler = PixelFormatNone
```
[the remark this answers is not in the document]
nveeser: done

channel_layout.go
```diff
@@ -123,3 +124,25 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
 	dst := newChannelLayoutFromC(&cl)
 	return dst, err
 }
+
+func ParseChannelLayout(s string) (ChannelLayout, error) {
```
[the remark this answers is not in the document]
nveeser: I paused here - I don't know the library that well but its seems to be both looking for an existing layout by name _and_ parsing a string which can be used to build a custom layout. 

Would it make more sense to use the terminology of the underlying library ChannelLayoutFromString()

channel_layout.go
```diff
@@ -123,3 +124,25 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
...
+
+func (l ChannelLayout) MarshalText() ([]byte, error) { return ([]byte)(l.String()), nil }
+func (l *ChannelLayout) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		return nil
+	}
+	pf, err := ParseChannelLayout(s)
+	if err != nil {
+		return fmt.Errorf("invalid Channel Layout: %s: %w", s, err)
```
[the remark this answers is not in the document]
nveeser: I wanna offer that adding the string to the context is pretty important (ala strconv.Atoi vs json.Marshal)
The only error would be EINVAL or ENOMEM w/ no logging so there is no other context. At least in the context of parsing JSON, if you are parsing multiple values in a single file, getting "invalid" as the only error is going to very difficult to debug.

Is the concern that including the argument string in the error value is too long?

log.go
```diff
@@ -22,6 +22,27 @@ const (
 	LogLevelDebug   = LogLevel(C.AV_LOG_DEBUG)
 )
 
+func (l LogLevel) String() string {
```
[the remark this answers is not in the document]
nveeser: sure - is there some precedent or motivation for this?

pixel_format.go
```diff
@@ -221,3 +228,17 @@ func FindPixelFormatByName(name string) PixelFormat {
 	defer C.free(unsafe.Pointer(cn))
 	return PixelFormat(C.av_get_pix_fmt(cn))
 }
+
+func (f PixelFormat) MarshalText() ([]byte, error) { return ([]byte)(f.String()), nil }
+func (f *PixelFormat) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		*f = PixelFormatNone
```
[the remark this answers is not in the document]
nveeser: oops thank you

pixel_format_test.go
```diff
@@ -7,10 +7,23 @@ import (
 )
 
 func TestPixelFormat(t *testing.T) {
-	p := FindPixelFormatByName("yuv420p")
-	require.Equal(t, PixelFormatYuv420P, p)
-	require.Equal(t, "yuv420p", p.String())
 	d := p.Descriptor()
 	require.NotNil(t, d)
 	require.Equal(t, d.Name(), p.String())
+	p := FindPixelFormatByName("yuv420p")
```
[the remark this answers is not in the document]
nveeser: I flattened things down so hopefully that makes it less redundant.

My goal was to separate the tests by at least the exported method. My assumption was that making it easy to read and reason about what all is suffiently tested and where is probably better than ensuring that there is no repitition anywhere? Does that seem reasonable?

pixel_format_test.go
```diff
@@ -7,10 +7,23 @@ import (
...
+	require.Equal(t, PixelFormatYuv420P, p)
+	require.Equal(t, "yuv420p", p.String())
+	t.Run("FindByName", func(t *testing.T) {
+		p := FindPixelFormatByName("yuv420p")
+		require.Equal(t, PixelFormatYuv420P, p)
+		require.Equal(t, "yuv420p", p.String())
+	})
+	t.Run("Encode", func(t *testing.T) {
```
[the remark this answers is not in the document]
nveeser: oops - yes

rational.go
nveeser: Yeah I think I tried json marshaling and then found cases where json did not work and encoding was the only option (cli flags or something).

I have seen in the past where having both was valuable but I cannot remember the context enough - so I just removed it as unecessary.

PR #166, opened 2025-09-17
ffmpeg version 8.0 with tested examples

oldma3095:


asticode (changes requested):
First off thanks for the PR ❤️ 

However I have 2 main concerns:

- I want this PR to be about n8.0 only therefore most of my comments are about removing whatever is not mandatory for n8.0. It doesn't mean those changes are not interesting, just that I would rather have them in other PRs.
- I don't want to change details in this PR, we really need to have the least changes possible in the only goal of being compatible with n8.0

oldma3095:
i need pull a better request

asticode:
Sure, bear in mind that for the "ffmpeg 8.0" PR, we're "just" looking for the least amount of changes so that tests pass when using the `ffmpeg n8.0` branch 👍

PR #171, opened 2025-11-06
Add metadata support to Frame

Tryanks:
I'm using cropdetect, but loss some feature...


frame_test.go
```diff
@@ -131,3 +131,19 @@ func TestFrame(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, f6b, f7b)
 }
+
+func TestFrameMetadata(t *testing.T) {
```
asticode: Can you move this logic at the bottom of `TestFrame` instead of creating a new `TestFrameMetadata`? You can rename `f` to `f8` in this case 👍

PR #173, opened 2025-11-29
Encoding: add encoding types for several enum types

nveeser:
Update Rational, PixelFormat and ChannelLayout to implement json interfaces which allow serializing these types to JSON without intermediate parser steps

Also add names for levels so the level name can be printed using fmt.Printf("%s")


nveeser:
My apologies for creating a separate branch / pull request.
I switch from git to jj and organized all the changes into a chain but I deleted a number of branches and re-exporting lost the history and therefore the PR was in a terminal state.

Apologies as the original comments are not here but instead on the original PR. 
I went through and tried to resolve all the comments in https://github.com/asticode/go-astiav/pull/161

I will try to copy over open questions here.

channel_layout.go
```diff
@@ -123,3 +130,28 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
 	dst := newChannelLayoutFromC(&cl)
 	return dst, err
 }
+
+// https://ffmpeg.org/doxygen/7.0/group__lavu__audio__channels.html#ga16a894cd275eba471ccad4fe1c5e5c28
+func ChannelLayoutFromString(s string) (ChannelLayout, error) {
```
nveeser: I don't know the library that well but its seems to be both looking for an existing layout by name and parsing a string which can be used to build a layout struct.

Since its both a "Find" and a "Parse", would it make more sense to use the terminology of the underlying library ChannelLayoutFromString()

channel_layout.go
```diff
@@ -123,3 +130,28 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
...
+	var c C.AVChannelLayout
+	cn := C.CString(s)
+	defer C.free(unsafe.Pointer(cn))
+	if err := newError(C.av_channel_layout_from_string(&c, cn)); err != nil {
+		return ChannelLayout{}, fmt.Errorf("astiav: could not find or parse channel layout %q: %w", s, err)
```
asticode: Could you replace the error message with `astiav: channel layout from string %q failed: %w`?

channel_layout.go
```diff
@@ -123,3 +130,28 @@ func (l ChannelLayout) clone() (ChannelLayout, error) {
...
+
+func (l ChannelLayout) MarshalText() ([]byte, error) { return ([]byte)(l.String()), nil }
+func (l *ChannelLayout) UnmarshalText(d []byte) error {
+	s := string(d)
+	if s == "" {
+		return nil
+	}
+	pf, err := ChannelLayoutFromString(s)
+	if err != nil {
+		return err
```
asticode: Could you return this instead `fmt.Errorf("astiav: finding channel layout failed: %w", err)`?

rational.go
```diff
@@ -2,13 +2,23 @@ package astiav
...
+	"strings"
+)
 
 // https://ffmpeg.org/doxygen/7.0/structAVRational.html
 type Rational struct {
 	c C.AVRational
 }
 
+var (
+	_ encoding.TextMarshaler   = (*Rational)(nil)
```
asticode: Can you replace with `_ encoding.TextMarshaler = NewRational(1, 1)`

asticode (changes requested):
No worries, I've answered in this new PR 👍

asticode:
I'll close this PR for now since it has conflicts but feel free to reopen it 👍

PR #175, opened 2025-12-04
fix io context

eTaulay:


asticode:
Thanks for the PR ❤️ 

Changes look fine but may I ask how did you find this (well hidden 🫣 ) bug? Is there a way to add a unit test for it?

eTaulay:
decode some audio, go prof discovered that io.ReadAll has a memory leak. Further investigation revealed that the readFunc closure was referencing unreleased memory

PR #183, opened 2026-05-12
feat: add calculation methods to rational

asix1337:
Add math helper methods to Rational using avutil/rational functions.


rational.go
```diff
@@ -53,3 +53,23 @@ func (r Rational) String() string {
 func (r Rational) Invert() Rational {
 	return NewRational(r.Den(), r.Num())
 }
+
+func (r Rational) Add(v Rational) Rational {
+	c := C.av_add_q(r.c, v.c)
```
asticode: Can you one line all below functions such as `return newRationalFromC(C.av_add_q(r.c, v.c))`?

rational.go
```diff
@@ -53,3 +53,23 @@ func (r Rational) String() string {
 func (r Rational) Invert() Rational {
 	return NewRational(r.Den(), r.Num())
 }
+
+func (r Rational) Add(v Rational) Rational {
```
asticode, line 58: Can you add links to C method doc at the bottom of each new function ?

Something like (for `av_add_q`)
```go
// https://ffmpeg.org/doxygen/8.0/group__lavu__math__rational.html#ga2eb3a275aabacd8421f140a12bab4a91
```

rational_test.go
```diff
@@ -17,6 +17,10 @@ func TestRational(t *testing.T) {
 	require.Equal(t, 2, r.Den())
 	require.Equal(t, "1/2", r.String())
 	require.Equal(t, 0.5, r.Float64())
+	require.Equal(t, "3/4", r.Add(NewRational(1, 4)).String())
```
asticode: Can you test using `.Float64()` instead of `.String()`?

asticode (changes requested):
Thanks for the PR ❤️ 

Just a few minor corrections needed 👍

asticode:
Thanks for the PR ❤️ 

PR #185, opened 2026-05-12
feat: add dictionary seq method

asix1337:
Add a Seq method to Dictionary to allow iterating over the key-value pairs. Needs at least go1.23.


asticode:
Thanks for the PR ❤️ 

However I'm not looking to add this to the lib since you can do the same with very few lines of code using the `.Get()` dictionary method that you use in this PR.

I'll close this PR 👍

PR #186, opened 2026-05-12
feat: expose long name of codec

asix1337:
Expose the "long_name" String.

### DIFF
--- a/codec.go
+++ b/codec.go
@@ -26,11 +26,7 @@ func (c *Codec) Name() string {
 
 // https://ffmpeg.org/doxygen/8.0/structAVCodec.html#acb4afcfce3b6541ed70aa7f21d507422
 func (c *Codec) LongName() string {
-	if c.c.long_name != nil {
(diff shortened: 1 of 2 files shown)


asticode (changes requested):
Thanks for the PR ❤️ 

Could you add a test too?

codec.go
```diff
@@ -24,6 +24,15 @@ func (c *Codec) Name() string {
 	return C.GoString(c.c.name)
 }
 
+// https://ffmpeg.org/doxygen/8.0/structAVCodec.html#acb4afcfce3b6541ed70aa7f21d507422
+func (c *Codec) LongName() string {
+	if c.c.long_name != nil {
```
asticode: Can you simplify with only `return C.GoString(c.c.long_name)`? (no need to check for `nil`)

asticode (changes requested):

asticode:
Thanks for the PR ❤️ 

PR #188, opened 2026-05-12
feat: expose probe score in format context

asix1337:
Add ProbeScore to FormatContext to get "probe_score".


asticode:
Thanks for the PR ❤️ 

PR #191, opened 2026-07-15
Add support for chapters

asix1337:
This pull request will add support for chapters. Tests and a mp4 file which includes chapters are also part of it.


asticode (changes requested):
Thanks for the PR! ❤️ 

Can you add tests in `format_context_test.go` as well for the 2 `FormatContext` methods you've added? No need to test `Chapter` structs fully for the `.Chapters()` method, but testing the length of the slice and the ids of the elements should be enough 👍

format_context_test.go
```diff
@@ -8,13 +8,13 @@ import (
 )
 
 func TestFormatContext(t *testing.T) {
-	fc1, err := globalHelper.inputFormatContext("video.mp4", nil)
+	fc1, err := globalHelper.inputFormatContext("video_with_chapters.mp4", nil)
```
asticode: How did you create `video_with_chapters`? Did you take the `video.mp4` in the `testdata` folder and added chapters manually using ffmpeg?

format_context.go
```diff
@@ -265,6 +279,25 @@ func (fc *FormatContext) CloseInput() {
 	}
 }
 
+// https://ffmpeg.org/doxygen/8.0/demux__utils_8c.html#a29bbc47c9d4d0f26439da347eba9a15a
```
asticode, line 282: Why not use the `avpriv_new_chapter()` method?

asix1337: Because avpriv_new_chapter() is not part of FFmpeg's public API. It's may not be API stable and we dont have to write a function with much parameters.

format_context_test.go
asix1337: Yes, using the following command: ffmpeg -i video.mp4 -i chapters.txt -map_metadata 1 -codec copy video_with_chapters.mp4. The file "chapters.txt": 
;FFMETADATA1

[CHAPTER]
TIMEBASE=1/1000
START=0
END=2500
title=Chapter 1 - First Half

[CHAPTER]
TIMEBASE=1/1000
START=2501
END=5000
title=Chapter 2 - Second Half

format_context.go
asix1337: C.avpriv_new_chapter() is also not accessible for me.

format_context_test.go
asticode: nice, in that case could you replace the `video.mp4` instead of creating a new file? And make sure the current tests still pass (or make the proper modifications so that they do)?

format_context.go
```diff
@@ -265,6 +279,25 @@ func (fc *FormatContext) CloseInput() {
...
+func (fc *FormatContext) NewChapter() (*Chapter, error) {
+	ch := (*C.AVChapter)(C.av_mallocz(C.size_t(unsafe.Sizeof(C.AVChapter{}))))
```
asticode: Can you use `C.sizeof_AVChapter` instead of `unsafe.Sizeof(C.AVChapter{})` here ?

format_context.go
```diff
@@ -265,6 +279,25 @@ func (fc *FormatContext) CloseInput() {
...
+	if ch == nil {
+		return nil, fmt.Errorf("astiav: allocation is nil")
+	}
+
+	ret := C.av_dynarray_add_nofree(
```
asticode: Can you "simplify" with
```go
if ret := ...; ret < 0 {
  ...
}
```

format_context.go
```diff
@@ -265,6 +279,25 @@ func (fc *FormatContext) CloseInput() {
...
+		unsafe.Pointer(&fc.c.chapters),
```
asticode: Do you have to use `unsafe.Pointer` in all lines here?

Or can you do instead
```go
C.av_dynarray_add_nofree(&fc.c.chapters, (*C.int)(&fc.c.nb_chapters), ch)
```

asix1337: unfortunately i have to use unsafe.Pointer.

format_context_test.go
asix1337: i tried to replace video.mp4 with video_with_chapters.mp4, but some tests like "TestFrame" fails though the added stream which is "bin_data" on "globalHelper.inputLastFrame("video.mp4", MediaTypeVideo, nil)". Maybe it's a better solution to hold both files.

asticode: in that case, can you avoid updating the file used here and instead open `video_with_chapters.mp4` at the bottom of `TestFormatContext()` just to test chapters?

Something like:
```go
fc10, err := globalHelper.inputFormatContext("video_with_chapters.mp4", nil)
require.NoError(t, err)
require.Equal(t, 2, fc10.NbChapters())
require.Len(t, fc10.Chapters(), 2)
```

asticode:
Thanks a lot for the PR! ❤️ 

asix1337:
> Thanks a lot for the PR! ❤️
> 
> Let me know whether you need a tag 👍

That would be nice ❤️

asticode:
> > Thanks a lot for the PR! ❤️
> > Let me know whether you need a tag 👍
> 
> That would be nice ❤️

FYI I' ve created a `v0.42.0` tag 👍

PR #192, opened 2026-07-16
fix warnings concerning golang

asix1337:
I removed some code which my golang checker told me to fix. I hope you accept it.


asticode:
Thanks for the PR! ❤️

PR #194, opened 2026-08-15
Add metadata example

asix1337:
This will add a metadata example to get some informations about streams, chapters and format.


asticode:
Thanks for the PR! ❤️

PR #197, opened 2026-09-19
Support for FFmpeg 9

njyeung:
ffmpeg 9 dropped `#include avutil.h` from `libavutil/opt.h`, which is what transitively declared av_freep (opt.h -> avutil.h -> common.h -> mem.h), so option.go no longer compiled.

Other files that call av_freep are still able to reach mem.h through their headers.

ffmpeg 9 also removed the V308/V408/V410 codec ids.

Compiles against ffmpeg 9.0.1 and 8.0.1. Tests were failing due to exact string assertions which I modified to pass.


asticode:
First off thanks for the PR ❤️ 

Since its goal is to make `go-astiav` compatible with ffmpeg 9:

1) we should target `n9.0`
2) you should update all references to `8.0` to `9.0` in the repo (that should either be doc links, README, github action conf or devcontainer conf)

That way I'll be able to run github actions against ffmpeg 9 and see whether all tests pass in every OS 👍

njyeung:
Should be good to run tests now 👍

I also ran them on my end: https://github.com/njyeung/go-astiav/actions/runs/35664654141

.github/workflows/test.yml
```diff
@@ -64,7 +64,6 @@ jobs:
       run: |
         choco install make
         choco install nasm
-        echo "FFMPEG_PATCH_PATH='$(cygpath -u ${{ github.WORKSPACE }})/.github/workflows/windows.patch'" >> $env:GITHUB_ENV
```
asticode: Why did you remove this? 🤔

.github/workflows/windows.patch
asticode: Why did you remove this? 🤔

option.go
```diff
@@ -1,13 +1,14 @@
...
+//#include <libavutil/mem.h>
 //#include <libavutil/opt.h>
 //#include "option.h"
 import "C"
 import (
 	"unsafe"
 )
 
-// https://www.ffmpeg.org/doxygen/7.0/structAVOption.html
+// https://ffmpeg.org/doxygen/9.0/structAVOption.html
```
asticode: Good catch! 👍

.github/workflows/test.yml
njyeung: This commit fixes the bug described in windows.patch:

https://github.com/FFmpeg/FFmpeg/commit/4ba56d8ab76308adb4c184ffb203933e3a28e1b3

asticode: Another good catch then 👍

asticode:
Thanks again for the PR! ❤️ 

FYI I've created a `v0.43.0` tag 👍

